PR #42, opened 2021-11-25
chore: fix storybook build

riccardo-forina:


riccardo-forina:
Not sure why Chromatic fails building this, it works just fine locally. I'll merge anyway.

PR #50, opened 2021-11-29
fix: handle more edge cases with the time range selector

riccardo-forina:


.github/workflows/chromatic.yml
```diff
@@ -41,5 +47,5 @@ jobs:
         # Options required for Chromatic's GitHub Action
         with:
           #👇 Chromatic projectToken, see https://storybook.js.org/tutorials/intro-to-storybook/react/en/deploy/ to obtain it
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          projectToken: 233963c78c8a
```
riccardo-forina, line 50: This is the only way I found to make the Chromatic action work for forks

PR #53, opened 2021-11-30
feat(Metrics): add metric KPIs

riccardo-forina:


riccardo-forina:
@jgiardino this should finalize all the UI elements required for the Metrics dashboard. The link to the storybook is in the actions above.

src/components/Metrics/Metrics-i18n.json
```diff
@@ -11,9 +12,10 @@
     "connection_attempt_rate_yaxis": "Connections/second",
     "log_size_per_partition": "Log size per partition",
     "log_size_per_partition_help_text": "Topic partition size is the log size for each partition in a selected topic. This metric enables you to assess the amount of data used in one or more topic partitions. To reduce the topic partition size, you can decrease the retention time or the retention size, or modify the cleanup policy as needed. You can also increase the number of partitions for the topic.",
-    "incoming_bytes_per_topic": "Incoming bytes per topic",
-    "outgoing_bytes_per_topic": "Outgoing bytes per topic",
-    "total_bytes": "Total bytes",
+    "all_topics": "all topics",
+    "incoming_bytes": "Incoming bytes ({{topic}})",
+    "outgoing_bytes": "Outgoing bytes ({{topic}})",
+    "total_bytes": "Bytes incoming and outgoing",
```
jgiardino, line 18: I see that you have the right text here, but it isn't showing up in the storybook example: 
image

jgiardino: Oh, nevermind. I just thought to check if I was looking at the latest storybook. I wasn't. 🙃

src/components/Metrics/Metrics-i18n.json
```diff
@@ -1,5 +1,6 @@
 {
   "metrics": {
+    "axis-label-bytes": "Bytes",
     "axis-label-time": "Time",
     "used_disk_space": "Used disk space",
     "used_disk_space_popover_header": "Used disk space chart",
```
jgiardino, line 6: I don't actually see this text in the popover for this chart 🤔  but is it possible to remove the text "chart" from all of the `_popover_header` strings? We should just repeat the name of the chart as the popover header.

riccardo-forina: Sure!

riccardo-forina: Done

riccardo-forina: Btw that string is used to label the Y axis of the bytes related charts in the topic metrics card.

I added a tool to Storybook to disable the translations so to see the keys directly. It works like this
Kapture 2021-11-30 at 17 57 23

jgiardino:
These updates look really amazing! 

I used [this view from chromatic](https://www.chromatic.com/build?appId=619e1921b3e38b003afeb2be&number=80) to review and provide comments. I either 

- accepted it, because it looked good
- provided comments with some suggestions, but did not deny since I'm fine with pushing any of my suggestions to a future PR
- did neither of the above because I'm still thinking about these use cases

I also later realized that my comments included line breaks, but chromatic didn't preserve them, so it probably makes some of my comments really hard to read. I pasted the one comment that has the most line breaks and is maybe the more challenging suggestion to process mentally...


For **Metrics:Topics just created (no metrics)**
This scenario is really stumping me. When we discussed this before, it sounded like the API will not provide data if the user selects filters that do not return data. Hence the message "adjust your filters". But for the situation where there isn't any data because they just started using the Kafka instance, this message is potentially really confusing. 

When we discussed this before, these were the main scenarios that came up where no data is returned: 

1. Scenario 1 - user selected a topic filter for which data existed but was deleted 4 hours ago, then sets time range to last 2 hours 
2. Scenario 2 - user had topics, but deleted them 4 hours ago, and sets time range to last 2 hours 
3. Scenario 3 - user just created topics and started using them and we're not getting data yet 

For scenarios 2 and 3, no topics currently exist. It's true that they could adjust filters for scenario 2 to see data in the past, but it's also true that topics don't exist at the current moment. And we have no way of knowing that data exists in the past. 

For scenario 1, the user has a topic filter selected. 

So I'm wondering if this would work and provide reasonable messaging most of the time: 
- Given that no topic filters are selected, when no data is returned (because either topics don't exist or we're still processing data from recently created topics), then display the following message: 
 - No topic data - Data will appear shortly after you start using topics. 
 - (displays with the tachometer icon) 
- Given that a topic filter is selected, when no data is returned, then display the following message: 
 - No topic data - Adjust your filters and try again 
 - (displays with the Search icon)

riccardo-forina:
Thanks @jgiardino, I updated the PR taking into account your feedback.

I also pushed a revised setup for the stories where I try to provide some description about what's being tested. You can see it clicking on the "Docs" tab in any of the Metrics' stories.

pmuir:
For the sample data - note that butes incoming and outgoing are totals, not rates, so they can't go down, only up, unless you've managed to reverse the second law of thermodynamics ;-)

pmuir:
Why are we smoothing the lines for client connections and disk space but not bytes?

pmuir:
All the various states look good.

pmuir:
Code all looks great.

The y axis labels aren't very consistent:

* Used disk space - Used disk space (description)
* Bytes incoming and outgoing - Bytes (unit)
* Connetion attempt rate - Connections/second (unit)
*

riccardo-forina:
> Why are we smoothing the lines for client connections and disk space but not bytes?

No reason in particular I think. What would your preference be?

pmuir:
> > Why are we smoothing the lines for client connections and disk space but not bytes?
> 
> No reason in particular I think. What would your preference be?

I *think* I prefer non-smoothed lines. Smoothing is very useful to spot trends in noisy data, but in this case there is no noise in the data.

riccardo-forina:
Makes sense. That's for all charts, right?

pmuir:
Yes, all charts.

jgiardino:
> The y axis labels aren't very consistent:
> 
> Used disk space - Used disk space (description)
> Bytes incoming and outgoing - Bytes (unit)
> Connetion attempt rate - Connections/second (unit)

Good catch! The y axis labels are meant to describe the unit. 
I _think_ these are the labels that we should update and use:

- Used disk space: "Bytes"
- Client connections: "Connections" (do we need "Client" for this?)

riccardo-forina:
I should have addressed all feedback, and added also the last updated info in the cards
Kapture 2021-12-02 at 12 37 38

I'd merge this, we can always tune it more later.

PR #60, opened 2021-12-08
feat: add shared components and create stories for empty state component

ajaypratap003:


src/components/AppServicesEmptyState/AppServicesEmptyState.tsx
```diff
@@ -0,0 +1,167 @@
...
+  };
+  buttonProps?: Omit<ButtonProps, "children"> & {
+    title?: string;
+    onClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+    "data-testid"?: string;
+  };
+  children?: React.ReactNode;
+};
+
+export const AppServicesEmptyState: React.FC<AppServicesEmptyStateProps> = ({
```
riccardo-forina, line 54: I really don’t like generic empty state components that take a bunch of props to set it up. It makes it very easy to break it by setting the wrong props, and it's hard to know what's available.

Better having a component for each empty state, with only the props they need (often none).

Check the empty states in the metrics for an example.

src/utils/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const getModalAppendTo = (): HTMLElement =>
```
riccardo-forina, line 1: This doesn't belong to this repo, it's an app-services-ui concern knowing there to put modals

src/index.ts
```diff
@@ -1 +1,4 @@
 export * from './components';
+export * from './kas';
+export * from './kafka';
+export * from './service-registry';
```
riccardo-forina, line 4: I don't think we need all this structure. A single folder for the components will work just fine. It's all about giving components proper names.

riccardo-forina (changes requested):

src/components/AppServicesEmptyState/AppServicesEmptyState.stories.tsx
```diff
@@ -0,0 +1,152 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import {
+  AppServicesEmptyState,
+  AppServicesEmptyStateVariant,
+} from "./AppServicesEmptyState";
+
+export default {
+  title: "Shared Components/EmptyState",
```
riccardo-forina, line 9: There is already an "Empty states" section in Syorybook, no need to come up with a new one

src/components/AppServicesEmptyState/AppServicesEmptyState.css
```diff
@@ -0,0 +1,3 @@
+.pf-c-empty-state.pf-m-xl {
+  --pf-c-empty-state__content--MaxWidth: 624px;
```
riccardo-forina, line 2: This should be checked against the smaller breakpoints. There is a button on the toolbar in Storybook to try different layouts

src/components/AppServicesErrorBoundary/AppServicesErrorBoundary.tsx
```diff
@@ -0,0 +1,75 @@
...
+      );
+    }
+    return this.props.children;
+  }
+}
+
+const ErrorBoundaryComponent = withRouter(
+  withTranslation()(AppServicesErrorBoundary)
+);
+export { ErrorBoundaryComponent as AppServicesErrorBoundary };
```
riccardo-forina, line 75: This component belong to the app-services-ui, and the jsx part should be it's own empty state

src/components/AppServicesPageNotFound/AppServicesPageNotFound.tsx
```diff
@@ -0,0 +1,36 @@
...
+  AppServicesEmptyStateVariant,
+} from "../../components";
+
+const AppServicesPageNotFound: React.FunctionComponent = () => {
+  const { t } = useTranslation();
+
+  function GoHomeBtn() {
+    const history = useHistory();
+    function handleClick() {
+      history.push("/");
```
riccardo-forina, line 16: This is absolutely not a concern of this repo, it belongs to app-services-ui

riccardo-forina (changes requested):
Don't move the tables yet. They need a complete refactoring.

Try to focus more on separating the concerns and less on doing big changes at a time.

src/index.ts
ajaypratap003: As we are planning to move all the visual components from different modules (kas-ui, kafka-ui, srs-ui etc) , so it would be better to have separate folder for each module and it will easy to find out the components.
Top level components folder is for shared components only.
We already have similar structure for Metrics . we need to move Metrics out side of components folder.
It will be more clear.

src/components/AppServicesEmptyState/AppServicesEmptyState.tsx
ajaypratap003: We need to enhance shared components for better reuse. But at this movements let it be as it is.

src/components/AppServicesErrorBoundary/AppServicesErrorBoundary.tsx
ajaypratap003: Agreed but this ErrorBoundary component is used in kas-ui for running standalone kas-ui. 
If we don't need then we can remove but it will not run wrap with app-services-ui.

src/components/AppServicesPageNotFound/AppServicesPageNotFound.tsx
ajaypratap003: Agreed. Similar to ErrorBoundary it's also using kas-ui for running kas-ui standalone. It will not wrap in app-services-ui app.
We have routing in kas-ui as well for running standalone.

src/components/AppServicesEmptyState/AppServicesEmptyState.css
ajaypratap003: This is existing css in kas-ui. This was added for some reason by someone.

src/components/AppServicesEmptyState/AppServicesEmptyState.stories.tsx
ajaypratap003: Agreed,  But I was thinking to keep in shared components category as it's shared component. 
Otherwise we can move. Let me know if need to move.

riccardo-forina: You just need to change how stories are grouped. Something like `title: "Empty States/App Services/Description of the state"`

src/components/AppServicesErrorBoundary/AppServicesErrorBoundary.tsx
riccardo-forina: Not sure we need error boundaries in standalone. Better have the app to fail quick, it makes debugging the problem easier.

src/components/AppServicesPageNotFound/AppServicesPageNotFound.tsx
riccardo-forina: kas-ui will have to inject its own callback. No biggie.

riccardo-forina (approved):

PR #62, opened 2021-12-09
feat : Add UI component for create instance modal

suyash-naithani:


src/components/CreateInstance/InstanceInfo.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </TextListItem>
+          </GridItem>
+        </Grid>
+        <Button
+          isSmall
+          isInline
+          variant={ButtonVariant.link}
+          style={{ marginTop: '20px' }}
+          onClick={() =>
+            qsContext.setActiveQuickStart &&
```
riccardo-forina, line 102: This callback should be provided by the consumer of app-services-ui. So there is no need to depend on the quickstart here.

riccardo-forina (commented):
Just a quick thing I noticed.

package.json
```diff
@@ -64,11 +64,13 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "4.132.2",
+    "@patternfly/quickstarts": "^1.3.0",
     "@patternfly/react-charts": "6.28.4",
     "@patternfly/react-core": "4.152.4",
     "@patternfly/react-icons": "4.11.14",
     "@patternfly/react-table": "4.29.58",
     "@rhoas/app-services-ui-shared": "0.15.2",
+    "@rhoas/kafka-management-sdk": "^0.18.0",
```
riccardo-forina, line 73: No sdks allowed!

.github/workflows/chromatic.yml
```diff
@@ -7,8 +7,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
```
ajaypratap003, line 10: is there any reason for removing pull_request branches ?

src/components/CreateInstance/CreateInstance.stories.tsx
```diff
@@ -0,0 +1,385 @@
...
+    name: {
+      value: "suyash-test",
+    },
+  },
+  formSubmitted: false,
+  cloudRegions: cloudRegions,
+  //hideModal:onCloseModal(),
+  //isModalOpen:isModalOpen
+};
+CreationInProgress.storyName = "Creation in progress";
```
riccardo-forina, line 156: Shouldn't the form be disabled while we are creating the instance?
Cancel and close button too.

riccardo-forina (commented):
Thank you, this is good progress! I left some comments, a few are trivial and others are about the overall design of the component.

If you can fix the trivial ones we can get this merged and then make another pass on this to address the bigger concerns.

jgiardino:
I know this PR is closed, but wasn't sure where the best place was to capture feedback. 

First, it's really great to see this component in the storybook! Thanks so much for adding this one! Here are some observations I made in no specific order:
- The order of the items in the navigation do not match the order they are listed in the Docs view. This makes it a little challenging to review the variations. 
- In the *Checking quota for instance* example, we don't know yet which regions are available to the user, since we don't know if they're getting a trial or standard instance yet and regions vary depending on type of instance they're getting. Currently the regions menu is showing options, but it should be disabled until we know what type they get as illustrated in [this example](https://docs.google.com/presentation/d/1IkRs3QVHwhTZvOny1LmVbNrEekMfyQuY_aYSZ9woYXU/edit#slide=id.g104ef8355c5_0_33). 
- Some of the examples don't show the final text and show the string id instead. Being able to see the actual text is really beneficial for some of us. This might already be something you're planning to address, but wanted to note it in case. 

This might be more general feedback about storybook, and less about the Create modal
- The order of the Canvas and Docs tabs have recently changed. It used to default to Canvas, and I could click the items in the left to view the canvas for that variation. Now, if I switch to the Canvas tab, that selection isn't persistent if I navigate to a different variation, and it switches to the Docs tab. I'm not sure really what's ideal behavior or what our options are, or if anyone else would find benefit in defaulting to the Canvas tab, but just wanted to share that feedback.

PR #64, opened 2021-12-10
feat: updated DeleteModal

riccardo-forina:


src/shared/AppServicesDeleteModal/AppServicesDeleteModal.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+      id={id}
+    >
+      <AppServicesDeleteModal
+        {...args}
+        isModalOpen={isModalOpen}
+        getModalAppendTo={() => document.getElementById(id)}
+        handleModalToggle={() => setIsModalOpen(false)}
+      />
+      <button onClick={() => setIsModalOpen(true)}>Open modal</button>
+    </div>
```
riccardo-forina, line 32: @suyash-naithani check how I handled the modal here, I think we have to have them start closed and then open manually for them to work peacefully with Storybook.

.storybook/main.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
     "@storybook/addon-a11y",
-    "storybook-i18n/preset",
```
riccardo-forina, line 12: I noticed this addon not working nice with the docs addon. I removed it and replaced its features in the `preview.js` directly.

src/shared/AppServicesDeleteModal/AppServicesDeleteModal.stories.tsx
riccardo-forina: I have actually expanded this quite a bit from when I pinged you. Please check the latest version.

riccardo-forina:
@ajaypratap003 I changed a bit the delete modal to make it customizable by using React nodes and fewer props, removing the configuration options that we don't need to make customizable since this modal is specific to the "delete something" domain.

@jgiardino can you check the "Components/Shared/Delete modal" stories and see if I missed something? I'm not sure about the inline alerts in case of errors when triggering the delete action. It's not a pattern documented in the design guidelines, but I feel like we need something like this. Showing a toast in case of error feels wrong, given that we are already in an overlay.
Another doubt, checking the design when there is a confirmation text field the warning icon is not present. Is that made on purpose? If so I have to change this a bit to accommodate that detail.

ajaypratap003 (approved):

PR #68, opened 2021-12-14
feat: centralize localizations

riccardo-forina:
## Changes
Provide a standard I18nProvider that can accept async resolvers to load the different json.
Make all translations namespaced.
Make the translation strings automatically extracted from the code at build time.

Unrelated to the i18n feature but required to fix the library, the code is now build and bundled with rollup in order to be able to bundle static assets like CSS and SVG files.

# What does this mean?

The copy we show to the user is as important as the UI itself. We want it to be checked, reviewed, and properly reused. We can build a glossary and refer to it in the translations, eg `"go_to_kafka_instances": "Go to $t(common:kafka) Instances"`.

## An easier development workflow

There is no need to update the json manually. The workflow will look like this:

1. do some changes that adds some translation key
2. run `npm run parse-i18n` to extract the strings and add them to the right namespaces
3. commit the changes

Eventually we'll add husky to the repo and make the keys extraction enforced and automatic.

## New rules

Every translation key should be explicitly part of a namespace, eg `t('common:kafka')` instead of `t('kafka')`.
Keys without a namespace will end in a `translations` namespace which feels like the rug where to hide the mess.
This rule is not yet enforced, but it should with an eslint rule.

## An easier copy review process

With all the translations in a single place it should be easier to check the copy in batches, and do cross-references as mentioned above (glossary).

Eventually, we'll add a page in Storybook with a table to show all keys and relative translations, to make it easier to spot missing translations or repetitions.

## One single translations provider

app-services-ui will have to use the `I18nProvider` component added in this PR, setting up with the proper strategy to load the locale json files also made available by this PR. Something along the lines of:

```ts
import { I18nProvider } from '@rhoas/app-services-ui-components';

<I18nProvider
  lng={'en'}
  resources={{
    en: {
      common: () => import('@rhoas/app-services-ui-components/locales/en/common.json'),
      'create-kafka-instance': () =>
        import('@rhoas/app-services-ui-components/locales/en/create-kafka-instance.json'),
      kafka: () => import('@rhoas/app-services-ui-components/locales/en/kafka.json'),
      metrics: () => import('@rhoas/app-services-ui-components/locales/en/metrics.json'),
    },
  }}
>
  ...
</I18nProvider>
```
Test commit: https://github.com/riccardo-forina/app-services-ui/commit/b5a59b28150e9280f9bb7a4922941b051f05562f

Note that imports must be statically analyzable by webpack to make code-splitting work, so we have to manually specify the imports one at a time.

The result is that the app will load the different json at boot time.
image

_Unfortunately, we can't load the namespaces on demand, react-i18next doesn't allow that to avoid flashing untranslated keys to the user._

*Very important*: federated modules *must not* create their own I18nProvider wrapper. Any instance of `useTranslation` will refer to the nearest provider. Having multiple of them means having multiple versions of the translations, which means we could show mismatched information, and also leads to unnecessary downloads for the user.


.storybook/preview.js
```diff
@@ -102,36 +99,46 @@ export const globalTypes = {
...
     description: "Internationalization locale",
-    defaultValue: "en_US",
+    defaultValue: "en-US",
     toolbar: {
       icon: "globe",
       items: [
-        { value: "en_US", right: "🇺🇸", title: "English" },
-        { value: "pseudo", right: "🤪", title: "Pseudo localization" },
+        { value: "en-US", right: "🇺🇸", title: "English (US)" },
+        { value: "en-GB", right: "🇬🇧", title: "English (GB)" },
```
jgiardino, line 107: So are we differentiating between US and GB english now? We'll need to find a translator for the GB English ;-P

jgiardino:
Your description of these updates sounds good to me!

.storybook/preview.js
riccardo-forina: Actually no! We only use the main part of the language (so, en) but this is handy to check that this mechanism works since browsers will give us the full language plus variant (so, en-GB)

riccardo-forina: Italian too, it is just to check that the fallback to en works.

riccardo-forina:
I forgot to mention, it's now possible to use markdown in the translation file to write basic markup like bold, italic, etc. without messing with the HTML.

PR #69, opened 2021-12-15
feat: Add story for development preview

HemaHG:
Add story for development preview


src/components/DevelopmentPreview/DevelopmentPreview.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+    },
+} as ComponentMeta<typeof DevelopmentPreview>;
+
+
+const Template: ComponentStory<typeof DevelopmentPreview> = (args) => (
+    <DevelopmentPreview {...args} />
+);
+
+export const showDevPreview = Template.bind({});
+showDevPreview.args = {};
```
riccardo-forina, line 22: I think this should show how this looks like with a children, like a sample PageSection or something like that.

It's to understand margins and how the component react at the different breakpoints.

PR #83, opened 2021-12-20
 feat: Add story for Overview page

HemaHG:
Add story for Overview page


src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,332 @@
...
+    };
+
+    const getServiceRegistryHref = () => {
+        if (beta) {
+            history.createHref({
+                pathname: '/service-registry',
+            });
+        }
+        return '/beta/application-services/service-registry';
+    };
```
riccardo-forina, line 66: All of these should be props passed to the component, how navigation works is an app-services-ui concern

locales/en/common.json
```diff
@@ -7,6 +7,37 @@
   "form_invalid_alert": "Address form errors to proceed.",
   "go_to_kafka_instances": "Go to $t(common:kafka) Instances",
   "kafka": "Kafka",
+  "overview": {
```
riccardo-forina, line 10: I'd move this to its own file. To do it just namespace the translations using : between the namespace and the key. Like:

`t('overview:alpha')`

Then run `npm run parse-i18n` to collect all the keys automatically.
A new overview.json file will be created, move the keys there (only those needed).

If a translation.json file gets created too, you have non-namespaced keys that have to be fixed.

src/OverviewPage/OverviewPage.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import React from "react";
+import { OverviewPage } from './OverviewPage'
+
+export default {
+    title: "Components/OverviewPage",
```
riccardo-forina, line 6:
```suggestion
    title: "Features/OverviewPage",
```
 Let's make this more prominent, it's not really a reusable component.

src/OverviewPage/OverviewPage.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+    component: OverviewPage,
+    args: {},
+} as ComponentMeta<typeof OverviewPage>;
+
+const Template: ComponentStory<typeof OverviewPage> = (args) => (
+    <OverviewPage {...args} />
+);
+
+export const Overview_page = Template.bind({});
```
riccardo-forina, line 15:
```suggestion
export const OverviewPage = Template.bind({});
```

The space gets added by storybook automatically using camel case in the const name

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+import APIMgmtLogo from './static/images/Logo-Red_Hat-OpenShift-API_Management-A-Standard-RGB.svg';
+import DataScienceLogo from './static/images/Logo-Red_Hat-OpenShift-Data_Science-A-Standard-RGB.svg';
+import ServiceRegistryLogo from './static/images/Logo-Red_Hat-OpenShift_Service_Registry-A-Standard-RGB.png';
+import DataServicesLogo from './static/images/Logo-Red_Hat-Data_Services-A-Standard-RGB.png';
+import './OverviewPage.css'
+
+import { useTranslation } from 'react-i18next';
+
+export type OverviewPageProps = {
+    onClickKafkainstance: (event: any) => void;
```
riccardo-forina, line 33:
```suggestion
```

Never use `any`. There is a rule about this too in the linter.

In this case there is no need to pass back the event, we already have the href, no need for anything else

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+    getKafkaHref: string;
+    onClickServiceRegistry: (event: any) => void;
```
riccardo-forina, line 35:
```suggestion
   
```

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+import DataScienceLogo from './static/images/Logo-Red_Hat-OpenShift-Data_Science-A-Standard-RGB.svg';
+import ServiceRegistryLogo from './static/images/Logo-Red_Hat-OpenShift_Service_Registry-A-Standard-RGB.png';
+import DataServicesLogo from './static/images/Logo-Red_Hat-Data_Services-A-Standard-RGB.png';
+import './OverviewPage.css'
+
+import { useTranslation } from 'react-i18next';
+
+export type OverviewPageProps = {
+    onClickKafkainstance: (event: any) => void;
+    getKafkaHref: string;
```
riccardo-forina, line 34:
```suggestion
    toKafkaHref: string;
```

`get` implies an action, I'd expect a function to call, not a primitive type

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+    onClickServiceRegistry: (event: any) => void;
+    getServiceRegistryHref: string;
```
riccardo-forina, line 36:
```suggestion
    toServiceRegistryHref: string;
```

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    <Title headingLevel="h1" size="2xl">
+                        {t('overview.heroTitle')}
+                    </Title>
+                    <Text className="app-services-ui--banner__tagline">{t('overview.heroTagline')}</Text>
+                    <Text component={TextVariants.p}>{t('overview.heroDescription')}</Text>
+                    <Text component={TextVariants.p} className="pf-u-mt-md">
+                        {t('overview.heroDescription2')}
+                    </Text>
+                    <StackItem>
+                        <Button
```
riccardo-forina, line 65: We want the link to generate a link that will work correctly with react router. This is not the way to do it.

Check the [documentation| https://www.patternfly.org/v4/components/button/#router-link]. Ignore the variant, just the component prop part.

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+                                    {t('overview.rhodsSecondaryText')}{' '}
+                                    <Button
+                                        data-testid="cardRHODS-linkOpenShift"
+                                        isInline
+                                        variant={ButtonVariant.link}
+                                        component="a"
+                                        target="_blank"
+                                        href="http://openshift.com"
+                                    >
+                                        {' '}
```
riccardo-forina, line 164: Is this space right?

src/OverviewPage/OverviewPage.tsx
```diff
@@ -0,0 +1,312 @@
...
+                            </Button>
+                        </CardFooter>
+                    </Card>
+                </Grid>
+            </PageSection>
+        </>
+    );
+};
+
+export default OverviewPage;
```
riccardo-forina, line 312: Why is this file indented with 4 spaces? The standard is 2. I understand we don't have husky in this repo yet, but please keep an eye open to these details

riccardo-forina (changes requested):

PR #107, opened 2022-01-07
chore: added instance creation popover

cfullam1:


jgiardino:
This is really exciting to see! 

One minor thing I noticed is that service account has a capital "S" ("Service account") but should be all lower-case in this context.

And seeing the results of the suggestion to include the Ready status as a node in the progress stepper highlights that it doesn't really fit there. This would be good to get some input on from other team members (we can start with sharing it with designers). But here are my thoughts on this:

- "Ready" isn't really a step in the creation process, it's the final result of the creation process. In other words, there are 3 steps during creation, and when step 3 of 3 is completed, it's ready. Something that really highlights this is the inconsistency in microcopy—all of the true steps end with "ing" because something is happening. There is no "ing" version of "Ready", it's just done at this point.
- I still agree with the idea behind showing Ready. I also like including the text "x of 4 steps completed". But if we are honest with how many steps there are (3) and show that, then there's a confusing disconnect between what "x of y steps" is communicating vs what the progress stepper is communicating. 
- I think a possible solution to support both ideas (showing the final result and showing "x of y steps") is for the progress stepper to provide an optional "completed" node that displays last, which is visually presented in a way that it clearly is not a step and instead just a summary of what the end result of all the steps are. That could be a possible enhancement to this component, but would require either support from PF or custom CSS, as well as a round of design work. Without discussing this with others, I don't know how critical or necessary this enhancement would be. It might be best to wait until we _know_ there is a usability issue with not providing this information. 
- Given the amount of work required with the previous bullet, these are some alternative options that are possible now:
    - Remove the text "x of 4 steps completed" and rely on the progress stepper to communicate completion progress
    - Remove the "Ready" node, since it isn't a step in creation.

src/PoCs/KafkaStatusPopover.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+export const KafkaStatusPopoverBody: VoidFunctionComponent<
+  KafkaStatusPopoverBodyProps
+> = ({ currentState }) => {
+  const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+  const currentStep = order.findIndex((v) => v === currentState);
+  console.log(currentStep);
+  return (
+    <div>
+      <p>This will be ready shortly. In the mean time you can create a Service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```
abigaeljamie, line 45:
```suggestion
      <p>This will be ready shortly. In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```

src/PoCs/PopoverFifteenMinutes.tsx
```diff
@@ -0,0 +1,100 @@
...
+  export const KafkaStatusPopoverBody: VoidFunctionComponent<
+    KafkaStatusPopoverBodyProps
+  > = ({ currentState }) => {
+    const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+    const currentStep = order.findIndex((v) => v === currentState);
+    console.log(currentStep);
+    return (
+      <div>
+        <Alert variant="warning" isInline isPlain title="This is taking longer than expected." />
+        <p>In the mean time you can create a Service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```
abigaeljamie, line 47:
```suggestion
        <p>In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```

src/PoCs/PopoverThirtyMinutes.tsx
```diff
@@ -0,0 +1,99 @@
...
+  export const KafkaStatusPopoverBody: VoidFunctionComponent<
+    KafkaStatusPopoverBodyProps
+  > = ({ currentState }) => {
+    const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+    const currentStep = order.findIndex((v) => v === currentState);
+    console.log(currentStep);
+    return (
+      <div>
+        <Alert variant="danger" isInline isPlain title="This is taking longer than expected." />
+        <p>In the mean time you can create a Service account under the <a href="http://www.google.com/">Connections tab. </a>If your Kafka instance continues to take longer than expected, open a <a href="http://www.google.com/">support case.</a> </p>
```
abigaeljamie, line 46:
```suggestion
        <p>In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab. </a>If your Kafka instance continues to take longer than expected, <a href="http://www.google.com/">open a support case.</a> </p>
```

abigaeljamie: I moved the hyperlink slightly so that "open a support case" is all hyperlinked. That way, if a screenreader reads the link out of context, it will still make sense to the user.

src/PoCs/KafkaStatusPopover.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+export const KafkaStatusPopoverBody: VoidFunctionComponent<
+  KafkaStatusPopoverBodyProps
+> = ({ currentState }) => {
+  const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+  const currentStep = order.findIndex((v) => v === currentState);
+  console.log(currentStep);
+  return (
+    <div>
+      <p>This will be ready shortly. In the meantime you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```
abigaeljamie, line 45:
```suggestion
      <p>This will be ready shortly. In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```

src/PoCs/PopoverFifteenMinutes.tsx
```diff
@@ -0,0 +1,100 @@
...
+  export const KafkaStatusPopoverBody: VoidFunctionComponent<
+    KafkaStatusPopoverBodyProps
+  > = ({ currentState }) => {
+    const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+    const currentStep = order.findIndex((v) => v === currentState);
+    console.log(currentStep);
+    return (
+      <div>
+        <Alert variant="warning" isInline isPlain title="This is taking longer than expected." />
+        <p>In the meantime you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```
abigaeljamie, line 47:
```suggestion
        <p>In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
```

src/PoCs/PopoverThirtyMinutes.tsx
```diff
@@ -0,0 +1,99 @@
...
+  export const KafkaStatusPopoverBody: VoidFunctionComponent<
+    KafkaStatusPopoverBodyProps
+  > = ({ currentState }) => {
+    const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+    const currentStep = order.findIndex((v) => v === currentState);
+    console.log(currentStep);
+    return (
+      <div>
+        <Alert variant="danger" isInline isPlain title="This is taking longer than expected." />
+        <p>In the meantime you can create a service account under the <a href="http://www.google.com/">Connections tab. </a>If your Kafka instance continues to take longer than expected,<a href="http://www.google.com/"> open a support case.</a> </p>
```
abigaeljamie, line 46:
```suggestion
        <p>In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab. </a>If your Kafka instance continues to take longer than expected,<a href="http://www.google.com/"> open a support case.</a> </p>
```

abigaeljamie (commented):
Some commas were missing in a couple of areas, so I just added those in with suggestions :)

kaedward (commented):
I believe "In the meantime, you can create a service account under the Connections tab." should be "In the meantime, you can create a service account under the Connections tab. If your Kafka instance continues to take longer than expected, open a support case."

src/PoCs/PopoverFifteenMinutes.tsx
kaedward, line 47: I believe "In the meantime, you can create a service account under the Connections tab." should be "In the meantime, you can create a service account under the Connections tab. If your Kafka instance continues to take longer than expected, open a support case."

kaedward: (Line 47)

cfullam1: Thank you fo looking over all of this @kaedward ! 
I think in one of the UX review meetings it was suggested to only link a support case after 30 minutes - I will double check this and if so, I will ensure it is noted in the design doc (it might be out of sync on this)!

src/PoCs/KafkaStatusPopover.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <br></br>
+      <p> {currentStep} of 4 steps completed</p>
+      <ProgressStepper isVertical>
+        <ProgressStep
+          variant={currentStep === 0 ? "info" : "success"}
+          isCurrent={currentState === "pending"}
+          description="Passed quota check and awaiting creation."
```
abigaeljamie, line 52: We don't need periods after these statuses because they're not sentences. The only one that is a sentence is the last one "Kafka instance is ready for use." We could shorten that so that it reads similarly to confirmations (like in a toast, even though this one isn't a toast). Example: "Kafka instance ready for use" or just "Ready for use"

cfullam1: @abigaeljamie @lboehling This is what the sample that I shared above would look like with two inline alerts (when the instance takes longer than 15 minutes to create).

Screenshot 2022-01-17 at 12 46 30

jgiardino:
Related to @lboehling's 3rd point above, the extra spacing is coming from the PF progress stepper. I created an issue in PF to address this: https://github.com/patternfly/patternfly/issues/4633

I would suggest we not try to address this spacing issue now, and wait for a fix to land in PF.

src/PoCs/threeStepPopover.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const KafkaStatusPopoverBody: VoidFunctionComponent<
+  KafkaStatusPopoverBodyProps
+> = ({ currentState }) => {
+  const order: States[] = ["pending", "provisioning", "preparing", "ready"];
+  const currentStep = order.findIndex((v) => v === currentState);
+  console.log(currentStep);
+  return (
+    <div>
+        <p>This will be ready shortly. In the meantime, you can create a service account under the <a href="http://www.google.com/">Connections tab.</a> </p>
+        <br></br>
```
jgiardino, line 65: Using line breaks to control spacing isn't ideal. I'm sure there's a good explanation on why somewhere out there—my best explanation is that there are a set of PF space sizes that are intended to create consistency with regard to space between UI elements. 

Ideally, we would have space applied to the `<p>` elements, something like `margin-bottom` set to `var(--pf-global--spacer--md)`.  And applying this space to the `<p> above and also below with the current step info would provide the spacing we need above the progress stepper.

Usually, you would put text contents into a [`<TextContent>` component](https://www.patternfly.org/v4/components/text), but that component is problematic inside the popover. In all of our popovers, we want the font-size to be consistent, but if we place a `<TextContent>` component inside the popover, then the font-size of the TextContent component will override the font-size of the popover making the text larger than expected. 

Some other options: 

- Use a [PF utility class](https://www.patternfly.org/v4/utilities/spacing) `pf-u-mb-md` on both `<p>` elements to apply `margin-bottom` using the `md` space size
- Create a new css class for this component that provides `margin-bottom: var(--pf-global--spacer--md);` and add the class to both `<p>` elements. There is probably a class name format that we would want to use for this, but I'm not sure what our standards for custom class names are.

@riccardo-forina, @srambach  - Do you have any recommendations for an approach to take here?

jgiardino:
Your latest version of this component looks really good! Thanks so much for working through all the open design questions in this PR.

Since we decided to keep the number of steps completed to keep alignment with a similar example in OpenShift (pasted below), I posted a comment inline about adding spacing for that. If we can match the spacing they have, that would be good, too.

image

The only other question I have is if we should move the KafkaStatusPopover story out of the POCS section and under the COMPONENTS section. I think the POCS was created while you were finalizing some of the design details, but I assume before we merge the PR, we want this in the expected location in the nav.

src/PoCs/threeStepPopover.tsx
srambach: I'd say unless this is truly a one-off, I'd recommend creating a custom style. Since the utility classes use `!important`, they are a little heavy-handed so if you can avoid them, it's not a bad idea.
I was wondering about using the `<TextContent>` with a class and setting a base font-size scoped to that, but I don't see an easy way of doing this.

PR #120, opened 2022-01-19
feat: Implement Search/Filter for service-account

HemaHG:
Implement Search/Filter for service-account
(https://issues.redhat.com/browse/MGDSTRM-4047)

Where to find stories
Components>ServiceAccount>Search


src/ServiceAccount/Search/ClientIDFilter.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const onChange = (input?: string) => {
+    setValue(input);
+    !valid && setValid(true);
+  };
+
+  const renderClientInput = () => {
+    if (filterSelected?.toLowerCase() === 'clientid') {
+      const v = !valid || isMaxFilter;
+
+      const FilterTooltip: React.FunctionComponent = () => {
```
riccardo-forina, line 63: It's better not to create a component inside a render function. I remember reading a good article about why this is bad, but I can't find it, unfortunately. In short, this doesn't do what you'd expect it to do, which is to create a React component that behaves like a React component.

Since the render function will be called by React every time it needs to render the React tree - which is pretty often, every time some state change somewhere - everything that's defined inside the render function will be evaluated and executed every single time. This means that your `FilterTooltip` component gets "reinvented" every time the `ClientIDFilter` one gets rendered. The result is that the previously rendered `FilterTooltip` when compared with the newly rendered `FilterTooltip` will look as a different component to React, making it render it every single time even if not needed.

The same is true for anything done inside the render function. Callbacks are another thing that can cause unneeded renders. This is why callbacks should be wrapped inside [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) unless they are static, meaning that they don't depend on anything coming as a prop.

src/ServiceAccount/Search/ClientIDFilter.tsx
```diff
@@ -0,0 +1,116 @@
...
+  onDeleteChip,
+  filterSelected,
+  isMaxFilter,
+  updateFilter,
+}) => {
+  const { t } = useTranslation();
+  const inputRef = useRef<HTMLInputElement>(null);
+  const tooltipContent = TooltipContent(isMaxFilter, 'clientid');
+
+  const [value, setValue] = useState<string | undefined>();
```
riccardo-forina, line 30: The `value` should be a prop, not a state of this component. https://overreacted.io/writing-resilient-components/#principle-1-dont-stop-the-data-flow

src/ServiceAccount/Search/ClientIDFilter.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [valid, setValid] = useState<boolean>(true);
```
riccardo-forina, line 31: This can be derived from the value of `value` at render time, it doesn't have to be saved as a state

riccardo-forina (commented):
I'm having some difficulties in reviewing this, there isn't much context about this PR to make me understand what was your goal.

When opening a PR you should always think about who is going to review it and if they can answer a few questions without reaching for further comments:
- what does the PR solve. This could be a link to an issue or a description you write while opening the PR
- how to check your work. It can be a screenshot, or in our case some guidance on how to find the stories on Storybook
- is there any work left / what's next

As a reviewer of the code, my focus is on checking if the React code is idiomatic and follows some principles to make it behave as a React component is expected to behave.
I think you'd benefit from going through [this article](https://overreacted.io/writing-resilient-components/) (I also linked this in a specific comment above).

As a user of Storybook, I don't find the stories clear enough to make clear their scope. I'm often at fault at this too. [Here is how to add a description to a story](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#add-description-to-individual-stories).

src/ServiceAccount/Search/FilterSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+    setIsFilterExpanded(!isFilterExpanded);
+  };
+
+  const onSelect = (
+    _event: React.MouseEvent | React.ChangeEvent,
+    selection: string | SelectOptionObject
+  ) => {
+    setIsFilterExpanded(!isFilterExpanded);
+    setFilterSelected(selection?.toString());
+  };
```
riccardo-forina, line 33:
```suggestion
  const onSelect = useCallback((
      _event: React.MouseEvent | React.ChangeEvent,
      selection: string | SelectOptionObject
    ) => {
      if (selection && typeof selection === 'string) {
        setIsFilterExpanded(false);
        onChange(selection);
      }
    },
    [onChange]
  );
```

riccardo-forina: Line 31: the type from the PF select for `selection` isn't great, so we need to do some checks on its value to let Tyepscript help us.

Line 32: you always want to close the dropdown when something gets selected. If you toggle the value of `isFilterExpanded`, you risk setting it `true` if something else for some weird reason changes it (which should not happen, but better safe than sorry).

Line 27 and 36: this needs to be wrapped in `useCallback`. The function depends on `onChange`, that's a property that is passed to the `FilterSelect` component and might change between renders. So we add it to the list of dependencies for `useCallback`

src/ServiceAccount/Search/FilterSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+    { label: t('kafka:short_description'), value: 'description', disabled: false },
+    { label: t('kafka:clientID'), value: 'clientid', disabled: false },
+    { label: t('kafka:owner'), value: 'owner', disabled: false },
+  ];
+
+  return (
+    <ToolbarItem>
+      <Select
+        variant={SelectVariant.single}
+        aria-label='Select filter'
```
riccardo-forina, line 45: `aria-label` needs a translated value too

src/ServiceAccount/Search/OwnerFilter.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onDeleteChip,
+  filterSelected,
+  isMaxFilter,
+  updateFilter,
+  value,
+  setValue
+}) => {
+  const { t } = useTranslation(
+
+  );
```
riccardo-forina, line 30:
```suggestion
  const { t } = useTranslation();
```

src/ServiceAccount/Search/OwnerFilter.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+  const tooltipContent = TooltipContent(isMaxFilter, 'owner');
+
+  const validate = (value?: string) => {
+    return value ? !/["$^<>|+%/;:,\s*=~#()]/.test(value.trim()) : true;
+  };
```
riccardo-forina, line 36: This function can be moved outside the render body.

riccardo-forina (commented):
I have some questions about this, let's review it together in a call. I left a few comments for the part I checked.

PR #123, opened 2022-01-22
feat: wire the create kafka instance dialog to its state machine

riccardo-forina:
I wrote a state machine to support the create Kafka instance dialog @suyash-naithani recently refactored.

I surely missed some state, could you check the stories?


jgiardino:
The updates you made to the Loading Data state, with showing the skeleton in the form and the info section, are really nice!

There are some missing messages or messages that should have different alert variants used, primarily related to the API checks we make after opening the modal vs on submit. We make similar API calls for each user action because it's possible that our info on loading the form has become stale. And we could get similar info for each different user action, but the message we display and how we display it will differ:

1.  **On loading the form**, any messages that communicate why a user can't create an instance display as **a warning alert, full width of the modal**. 
2. **On submitting the form**, any messages that communicate why a user can't create an instance display as **an error alert, spanning the width of the form only**.

Between cases 1 and 2, the message text will also change slightly

1. If the organization reached it's quota limit, these messages display:
    1. on form load, a warning alert displays: _Kafka instances are unavailable for creation / Your organization reached the limit of Kafka instances available for creation._
    2. on form submit, an error alert displays: _Kafka instance can't be created / Your organization reached the limit of Kafka instances available for creation._ 
2. If trial instances are unavailable, these messages display:
    1. on form load, a warning alert displays: _Kafka instances are unavailable for creation / Kafka instances might become available later, so check back in a few hours._
    2. on form submit, an error alert displays: _Kafka instance can't be created / Kafka instances are no longer available. Try again in a few hours._ 

Given these extra variations, it might help to update the story names to include whether the check is on form load or on form submit. This is an example of what that list could be:

- Over quota on form load
- Over quota on form submit
- Trial unavailable on form load
- Trial unavailable on form submit
- Trial used on form load (I guess in theory the user could create a trial instance in the CLI while simultaneously using the UI, but we never designed for that use case)

Given all the various messages that can display and when, I tried to capture details about message text, alert variant, etc... in a diagram [in these slides](https://docs.google.com/presentation/d/1IkRs3QVHwhTZvOny1LmVbNrEekMfyQuY_aYSZ9woYXU/edit#slide=id.g108e42fe94f_0_39). I also double-checked these slides to confirm they are correct, and made some updates today while reviewing your PR so this should be up-to-date with the latest text strings and color-coded to convey alert variant. 

Let me know if you need more specific information about what message and alert variant display for each state. I'm happy to outline all the expected details for the stories I listed above if that makes it easier to reference.


Aside from that, here are a couple more things I noticed:

1. There is a difference between what I had noted in the design document vs what we have currently implemented regarding when the submit button is disabled. For example, in the **Ready to create...** stories, my expectation is that the Create instance button would be enabled, but that isn't what we have implemented in the product today. My reason for keeping it enabled even when there are empty fields is that it might not be obvious to the user why the button is disabled if there is nothing flagged in the form. If the Create instance button in enabled, then they can click it and see a form-level error that explains why they can’t proceed along with field-level errors that highlight the things in the form they need to address.  I'm totally fine with revisiting this as a separate discussion/issue, thought.
2. In the **Generic error** story, the success state displays for Name and Cloud region. We don't have this in the Create instance modal today, but I have seen these used in some of our other modals in production and we should eventually refactor these to not have them. This success indicator should only be used when additional confirmation/validation is needed for the value provided by the user.
image

riccardo-forina:
Thank you Jenn, this is extremely useful. I updated the PR with the help of @suyash-naithani to cleanup the messages and reorganize the stories as you suggested. I still need to do the changes to support the always enabled Create button.

I have a question about this bit:

> We make similar API calls for each user action

This took me by surprise. Checking the existing implementation, I don't see API calls performed when say, the user selects a region. We load the data once when loading the modal, and then invoke the creation action which will either be successful or not.
Should we call the API more often? Maybe even poll for the data every X seconds?

jgiardino:
> We make similar API calls for each user action

What I meant was that we make API calls when the user opens the modal, and again when the user submits the form. Now that you mention it, your description sounds more accurate that how I describe it though.  I forgot that the on submit action is actually requesting a new Kafka, and we just present any errors from that action in the modal.

riccardo-forina:
This should be done now and it's ready to be reviewed (again).

* Form submit is always enabled
* Extended stories to show support for multiple providers
* Extended stories to show support for tweaking the defaults and the available options for the AZ

jgiardino:
You new set of stories look really good! I went through all of them and listed some miscellaneous things I noticed below.

Earlier today you asked about whether the required indicator is needed for Cloud provider, even when there's a default selection and you can't select it. I shared this question with other designers, and the consensus was that it should include the red *.

**Name Taken** - Today the error message for name that you show in this example displays only after the user submits the form. In our current implementation, an additional inline error alert like you have for *Form Errors* would also display.

Example of **Name Taken** in the latest storybook:
image

Example of current implementation:
image

**Trial Unavailable On Form Load** and **Trial Unavailable On Form Submit** - Both of these should include the "Duration / 48 hours" under _Instance information_:
image

**Only Single AZ Enabled With Right Tooltip** - I'm not sure if this use case will ever be needed, _but_ I would expect this to be similar to our use case today of only multi availability zones being supported

This is how it displays on load:
image

And then I can click the Single option to select it:
image

I think this just needs to have Single selected by default. The same tooltip should also display for the disabled Multi option. 

**No Defaults Mark Provider As Required** - Both options are enabled, but the tooltip still displays. In this case when no options are disabled, then the tooltip shouldn't display.

image

riccardo-forina:
> Earlier today you asked about whether the required indicator is needed for Cloud provider, even when there's a default selection and you can't select it. I shared this question with other designers, and the consensus was that it should include the red *.

Perfect, it's now on in all scenarios.

> **Name Taken** - Today the error message for name that you show in this example displays only after the user submits the form. In our current implementation, an additional inline error alert like you have for _Form Errors_ would also display.

Updated to work as expected.

> **Trial Unavailable On Form Load** and **Trial Unavailable On Form Submit** - Both of these should include the "Duration / 48 hours" under _Instance information_: image

Good catch, I had the stories set up wrongly. They should work as expected now. Doing this I noticed that in the `on Form Submit` state we will show two alerts, the top one relative to the instance lasting 48h and the form one to inform of the temporary unavailability of trial instances. I think this is correct, WDYT?
 
> **Only Single AZ Enabled With Right Tooltip** - I'm not sure if this use case will ever be needed, _but_ I would expect this to be similar to our use case today of only multi availability zones being supported
> I think this just needs to have Single selected by default. The same tooltip should also display for the disabled Multi option.

Exactly, I forgot to set the right default. 

> **No Defaults Mark Provider As Required** - Both options are enabled, but the tooltip still displays. In this case when no options are disabled, then the tooltip shouldn't display.

Good catch, there were a few issues with the way I was dealing with this. I also noticed that both the Region and the AZ options were enabled even without a provider selected. They are now disabled until you select a provider. As you said, it's not something we expect to happen but it's good to have it covered.

PR #132, opened 2022-01-28
chore:  Status Section for Kafka Table 

cfullam1:
WIP


cfullam1:
Hi @srambach I have been working on adding the status components for the Kafka instance table to storybook (https://issues.redhat.com/browse/MGDSTRM-7231). I have some questions and @jgiardino mentioned that you might have some insights:)

We would like for the icon(or spinner) and status not to be separated when the screen is made smaller. Currently this happens:
Screenshot 2022-02-02 at 12 56 50
but we would like for them to remain beside each-other.
I have been using flex layout in this pr (I also locally tried making the spinner and Creating text into one flex item but was met with the same results, them wrapping on separate lines). I am wondering whether you might have any thoughts on this wrapping issue?

@jgiardino also spotted that the spinner and icon e.g. check- circle icon  beside ready  are not strictly the same size and therefore the text "Creating" which switches to "ready" would not start in the same place.
Screenshot 2022-02-02 at 12 50 33
I am wondering whether using a custom size spinner is the appropriate resolution for this?  
Thank you!

cfullam1:
Hi @lboehling  I have made this pr for adding the status components for the Kafka instance table to storybook (https://issues.redhat.com/browse/MGDSTRM-7231). 

I remember that we decided that the secondary information would be in line with the Creating text land not the spinner like this. 
Screenshot 2022-02-02 at 13 09 58

I was looking to clarify, in the case of inline alerts, should they also start in the same place
Screenshot 2022-02-02 at 13 10 38

 or should they be in line with the beginning of the spinner? 
Screenshot 2022-02-02 at 13 12 09

If there are any other presentation/layout issues that you spot I will look into them too! Thanks!:)

srambach:
Hi @cfullam1 What do you think about using a split layout instead of a flex layout for the icon/spinner and text? Then it should not wrap. 
What size is your icon? The spinners are built with the same standard sizes as the icons, but if you don't have a standard icon size there, then you could certainly use the custom SVG spinner to make it match.

cfullam1:
Thank you @srambach! I will try it with the split layout. When I use a stand-alone icon it is 16x16 (I didn't specify this, it seems to be some sort of default) on the Patternfly site I didn't spot guidelines on implementing an icon (like react examples) but I think I saw that it should be 18x18, is there a specific way that this should be implemented? (The icons that come with the inline alerts also seem to be 18x18).

srambach:
Ah, you are correct - react-icons sizes the icons based on ems and not our icon sizes. I agree this is not clear in the documentation. You can see the sizing here in [react-icons](https://github.com/patternfly/patternfly-react/blob/ab8881705440faa1a8f31a49214067790be5c228/packages/react-icons/src/createIcon.tsx#L10-L23)
Take a look at the [progress button](https://www.patternfly.org/v4/components/button/html-demos#progress-button---loading). This is similar to what you want to do. The way it's done in that button is to use opacity to hide the icon so the button maintains its original shape, and then absolutely position the spinner in the middle. You could do a similar thing.

lboehling:
Hi @cfullam1! 

I'd suggest aligning the alert to the beginning of the spinner as you show below. This aligns the alert copy with the "creating" copy above. Aligning text to the left increases readability because it helps avoid unnecessary eye jumps. It also looks a bit neater IMO :)  I'd suggest keeping that text alignment when wrapping too, if possible (so no text would wrap below the alert icon and spinner). I think that's how PF alert titles are wrapped by default. 
> [](https://user-images.githubusercontent.com/89915947/152198262-07ded1b5-6ad2-4890-9345-e3047b984070.png)or should they be in line with the beginning of the spinner? 
image

cfullam1:
@srambach those links are helpful thank you! Just to clarify is the idea based on the progress button to horizontally align the spinner with the text?

cfullam1:
Thank you for this feedback @lboehling:) This layout/ wrapping sounds good to me.

srambach:
Hi, @cfullam1 I may have misunderstood; if you are trying to overlay the spinner and icon in the same spot, the progress button is doing it by absolutely positioning the spinner over the icon and then adjusting the opacity of the icon to hide it. 

However, if instead you just want to get things to line up nicely, you can make a few tweaks for that. Here's a codesandbox that shows using the split layout with an adjustment to the gutter size variable to align with the alert. It also shows how to adjust the icon size to be the standard 18px, so it's the same size as the spinner. 

Then there's also an example of adjusting the spinner down to the 16px, but I suspect you might want to adjust the icons up instead since the alert is using the standard 18px icon size.

https://codesandbox.io/s/icon-layouts-6j0uq?file=/index.tsx

cfullam1:
Hi @srambach, yes exactly it was to line things up and adjusting the icons up is ideal. Thank you for all of this help and for what you shared on codesandbox!

riccardo-forina:
Chiming in to say that I opened a PR that links the status label with the popover, ready to include the work from this PoC.
You can see the stories here https://619e1921b3e38b003afeb2be-xlmqydrlqn.chromatic.com/?path=/story/components-kafkainstancestatus--accepted or in the linked PR for the latest Storybook's build.

One thing I noticed that might Interest you is that the flex item needs to be displayed as an `inline-flex` for the popover to correctly position itself. That's if we don't want to set a ref object to the label, at least.
Also a question, what should be the action that triggers the popover? A click? On which part of the status? All of it, so the two lines and the icon? Just the top line?

cfullam1:
Thanks @riccardo-forina ! I think that the popover would be triggered by a click on only on the top line (i.e. the word "Creating"). WDYT @jgiardino? 
We have switched to using split layout in order to keep the icon or spinner beside each other in all cases (instead of going on different lines when the window size is reduced). I'm not sure how this would impact things i.e. would we need to still use an inline-flex for the popover to correctly position itself.

riccardo-forina:
How do we hint that's clickable? It wouldn't be obvious to me (I think, not sure)

cfullam1:
I think the button link component like this was the idea:
Screenshot 2022-02-04 at 13 05 34
We thought that the blue colour would show that it was clickable but maybe it's not clear and something to look into. WDYT @riccardo-forina ?

riccardo-forina:
Sure that's works. I tried using that in the linked PR and noticed that if you check the various states on next another the spacing is different between those with a popover and those without, since buttons have a minimum size to make them clickable and touchable.

cfullam1:
@riccardo-forina I added a change earlier today using the split layout and specified gutter sizing to have it match so hopefully the different statuses (button or not) line up now in the most recent commit. If it still looks off I can check it out!

jgiardino:
> I think that the popover would be triggered by a click on only on the top line (i.e. the word "Creating"). WDYT @jgiardino?

That's what I was thinking, too.

> We thought that the blue colour would show that it was clickable but maybe it's not clear and something to look into.

This would be a good thing to check in a follow-up usability test, since we're using PF defaults for styling.

src/PoCs/failed.tsx
```diff
@@ -0,0 +1,15 @@
...
+import "./style.css";
+
+
+export const InstanceCreationFailed: VoidFunctionComponent  = () => {
+    return (
+        <div>       
+            <Label color="red" icon={<ExclamationCircleIcon />}>
+                Failed
+            </Label>    
+        </div>
```
jgiardino, line 13: We shouldn't use the `<Label>` component for this status. Instead, the same  structure that you use for the Ready status can be used here.

src/PoCs/ready.tsx
```diff
@@ -0,0 +1,18 @@
+import { Split, SplitItem } from "@patternfly/react-core";
+import React, {VoidFunctionComponent } from "react";
+import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-icon';
+import "./style.css";
+export const ReadyForUse: VoidFunctionComponent  = () => {
+    return (
+        <div>       
+            <Split hasGutter className="foo">
+                <SplitItem>
+                    <CheckCircleIcon className="fixIcon" color="#3E8635" />
```
jgiardino, line 10: It would be good to use a common naming pattern for class names that we define.  I realize that the classes needed here are minimal, but establishing a pattern that we always follow will help when we get to components that require more customization via CSS.

Here's an example of the [class name that PF uses for their alert component](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Alert/alert.scss#L188): 
`.pf-c-alert__icon`

Following PF naming patterns, maybe something like `mas-c-status__icon` would work? 

And then `foo` could be changed to `mas-c-status`.

@riccardo-forina WDYT about using `mas-` as a class namespace?

Also, can colors be defined using PF color variables? Should color styles be moved to the CSS file? E.g. we could define the PF color variables for modifier classes like `mas-m-ready` and `mas-m-failed`

jgiardino:
Your latest updates look great! I posted a couple of inline comments with some questions and suggestions for managing the custom CSS.

The only other thing I noticed is the order of stories. I pasted below an example of the Docs view, where the first topic is about creation, but then the story that displays is Deleting. It would be good to fix this, but I was also wondering if the names and order should change a bit. What do you think about adopting this order and naming scheme?:

1. Ready
2. Creating (<15 minutes)
3. Creating (>15 minutes)
4. Creating (>30 minutes)
5. Deleting
6. Failed

image

src/PoCs/ready.tsx
riccardo-forina: > @riccardo-forina WDYT about using `mas-` as a class namespace?

Ah, it's a good idea, I have been using `appservices` so far. Your suggestion is shorter.
But I was thinking about making the class name not needed at all by adopting [CSS modules](https://css-tricks.com/css-modules-part-1-need/). We are not as good as UX at writing CSS nor we need to write that much CSS, so I think not having to deal with figuring out a class name by leveraging the automatic class names from CSS modules sounds like a good idea to me. WDYT?

riccardo-forina: > Also, can colors be defined using PF color variables?

I think so? In Storybook we have access to everything that [gets defined from these CSSs](https://github.com/redhat-developer/app-services-ui-components/blob/main/.storybook/preview.js#L1-L9). I think we'll have similar declarations on console dot.

We need to figure this out for certain.

jgiardino: > But I was thinking about making the class name not needed at all by adopting [CSS modules](https://css-tricks.com/css-modules-part-1-need/). We are not as good as UX at writing CSS nor we need to write that much CSS, so I think not having to deal with figuring out a class name by leveraging the automatic class names from CSS modules sounds like a good idea to me. WDYT?

I'm not familiar enough with how CSS modules work to know if this meets our needs. Ideally, whenever we extend PF components or layouts with custom CSS, we would use PF variables for all of our styles where they exist (e.g. color, font-size, spaces). This way if the PF visual theme is ever updated (e.g. the make one of the colors brighter, or modify relative font-sizes), then it's less likely to be a breaking change.

riccardo-forina: In short what CSS modules give you is a name for your classes that's automatically generated at build time. You don't need to come up with names that don't conflict with other existing classes and it's easier to refer to the CSS from the JS code.

Inside a CSS module you can still do everything you do with normal CSS files, so you have access to all variables etc.

CSS modules are good for one-shot customizations that you don't need to make reusable. They would not make sense when creating a CSS framework like PatternFly.

src/PoCs/deleting.tsx
```diff
@@ -0,0 +1,9 @@
+import React, {VoidFunctionComponent } from "react";
+
+export const InstanceDeletion: VoidFunctionComponent  = () => {
+    return (
+        <div>
+            <p style={{ color: "#6A6E73" }}>Deleting</p>
+        </div>
+    );
+};
```
jgiardino, line 9: The only question I have left is whether this color should also be moved to the CSS file and use the PF color variable instead.

jgiardino:
> Also that's strange it doesn't show up like that screenshot (with deleting first ) locally for me, but it switches to alphabetical order when I go to storybook through the link in the pr (I will look into this)

That is strange, and I forgot to look at this before I commented. This is what I see now, with Creating as the first story, which doesn't match the opening documentation.

@riccardo-forina What would you recommend in this case? Should we not try to include documentation at the beginning like we are currently in [ready.tsx](https://github.com/redhat-developer/app-services-ui-components/pull/132/files#diff-9c2e940558168a969f4144ccf6129e24fd102932ab85e2e8293d3244819023ecR7)? Another suggestion is to keep the extra intro documentation, but make it higher-level, and not try to describe the first component. 🤷‍♀️ 

image

riccardo-forina:
> I'm guessing this would be a separate follow-up task to this PR

Correct

riccardo-forina:
> > Also that's strange it doesn't show up like that screenshot (with deleting first ) locally for me, but it switches to alphabetical order when I go to storybook through the link in the pr (I will look into this)
> 
> 
> 
> That is strange, and I forgot to look at this before I commented. This is what I see now, with Creating as the first story, which doesn't match the opening documentation.
> 
> 
> 
> @riccardo-forina What would you recommend in this case? Should we not try to include documentation at the beginning like we are currently in [ready.tsx](https://github.com/redhat-developer/app-services-ui-components/pull/132/files#diff-9c2e940558168a969f4144ccf6129e24fd102932ab85e2e8293d3244819023ecR7)? Another suggestion is to keep the extra intro documentation, but make it higher-level, and not try to describe the first component. 🤷‍♀️ 
> 
> 
> 
> image
> 
> 

Stories are sorted by order of writing.

A workaround is to have an "example" story at the very top that's used to be a companion of the description/introduction that comes from the jsdoc of the component itself. Then have all the proper stories with their descriptions.

It's what I have done in the other PR about the status popover.

cfullam1:
> 

Thank you @riccardo-forina ! I've had this included at the top in the place of an example story (if this is correct) but it still switches to alphabetical order when I am not viewing locally. 
Screenshot 2022-02-09 at 12 55 36

I found this discussion  which seems to match my problem (https://github.com/storybookjs/storybook/issues/15574).

When you mentioned it would appear in the order it was written I realised that I had been organising things differently. I have been using separate files for each version of the story (ready, failed etc) I just noticed that you had them all in the same file. I should probably organise things like that too now and also this should fix the alphabetical issue!

riccardo-forina:
Ah, yes! The order of writing is for stories in the same stories file. For sibling stories files it's alphabetical.

PR #134, opened 2022-01-28
feat: menu component to add permissions to a Kafka instance

riccardo-forina:


riccardo-forina:
@jgiardino I'm trying to replicate the menu we need to build for the ACL shortcuts, but I'm not sure how to get it to work as we want it to.

You can preview the component under Components>ManageKafkaPermissions>PermissionsDropdown in Storybook

## Menu width
It looks like the width of the menu depends on the width of the menu toggle. To make the menu bigger, we also need to make the toggle bigger. I haven't found an escape hatch from this behavior.

## Descriptions
Words are poorly broken. This happens because of this CSS
```css
.pf-c-menu__item-description {
  font-size: var(--pf-c-menu__item-description--FontSize);
  color: var(--pf-c-menu__item-description--Color);
  word-break: break-all;
}
```
Setting `word-break` to `normal` gives something that looks better.
image

CC @ajaypratap003

riccardo-forina:
By the way, I followed the documentation [here](https://www.patternfly.org/v4/demos/composable-menu/) to implement this. I tried copy-pasting our copy in the previews into the documentation and the behavior is the same.

riccardo-forina:
Ok I found that specifying `popperMatchesTriggerWidth` to `false` to the `Popper` element we can detach the width of the menu from the width of the menu toggle. Doing this the menu will take all the space it needs to show the descriptions in a single line.

Would this be an option?

You can see both behaviours in the two stories, fixed width and auto width.

jgiardino:
The auto width option looks good! I am curious how this would work in the context of a modal on a page where the user has a smaller viewport. For example, should we set a max-width on it to be something like `calc(100vw - [the variable used for left/right padding *2])`? But I'm also fine with that being reviewed and addressed a separate PR.

riccardo-forina:
I honestly don't know how much we can control through CSS. 
All the sizing is computed through JavaScript by popper.js, I think. I don't see much in the documentation in terms of customization.

I’d keep the thing as it is and see where it takes us.

PR #142, opened 2022-02-02
feat : add instance status popover vertical variant

ajaypratap003:
https://issues.redhat.com/browse/MGDSTRM-7230


jgiardino:
Thanks for working on this, @ajaypratap003! I pasted below a screencap showing the latest updates in this PR and the original implementation (under POCS) and included a couple of things I noticed.

1. The text that was included in the original implementation has been modified. Can you use the text that was included in the original implementation? We were able to get content designers to review the contents in the initial PR, so that text should be final.
2. There are two examples missing from the original implementation: Error and Warning. Can you include those examples, too?
3. This popover is specific to Kafka. Should we keep the original label "KafkaStatusPopover" to avoid any confusion about which service this is for?
4. I missed this when reviewing the original implementation, but should we change the order in the nav to match the order of the steps (i.e. move Preparing Status after Provisioning Status)?

image

5. @cfullam1 had also included some additional documentation that displays in the Docs view shown below, but I don't see that in the Docs view of the new component. Is it possible to add this back in?
image

riccardo-forina:
One thing I noticed while quickly checking this, all copy should be under i18n.
I'll do a more in depth review tomorrow

ajaypratap003:
> Thanks for working on this, @ajaypratap003! I pasted below a screencap showing the latest updates in this PR and the original implementation (under POCS) and included a couple of things I noticed.
> 
> 1. The text that was included in the original implementation has been modified. Can you use the text that was included in the original implementation? We were able to get content designers to review the contents in the initial PR, so that text should be final.
> 2. There are two examples missing from the original implementation: Error and Warning. Can you include those examples, too?
> 3. This popover is specific to Kafka. Should we keep the original label "KafkaStatusPopover" to avoid any confusion about which service this is for?
> 4. I missed this when reviewing the original implementation, but should we change the order in the nav to match the order of the steps (i.e. move Preparing Status after Provisioning Status)?
> 
> image
> 
> 5. @cfullam1 had also included some additional documentation that displays in the Docs view shown below, but I don't see that in the Docs view of the new component. Is it possible to add this back in?
>    image

@jgiardino  I've added your suggestion. Please have a look and let me know if anything missing.

ajaypratap003:
> One thing I noticed while quickly checking this, all copy should be under i18n. I'll do a more in depth review tomorrow

@riccardo-forina Sorry, I don't get your point. Are you talking about translation strings ?

cfullam1 (commented):
From as far as I can see this is visually the same as the PoCs version now:)

src/KafkaInstanceStatusPopover/KafkaInstanceStatusPopover.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+   (Provides a link to the “Connections tab" where this is done).
+
+  -The steps shown in the progress stepper are statuses that show in the api. They have been grouped together under the one overall
+  “Creating status” in the table.There is some descriptive text accompanying each step too. When these steps are complete the instance will be ready for use
+
+  -The popover is only available while instance creation is in progress.
+
+  -After creation is complete the table displays the ready status and the popover is no longer available.
+
+  Pending Status
```
jgiardino, line 35: @riccardo-forina - do you know if we can use md to format the contents in this file, so that the dashes "-" display as bullets and the "Pending Status" can display as a heading?

jgiardino:
Thanks for making the updates, @ajaypratap003. Everything looks good!

Reviewing these examples, I had a couple of questions that I wanted to raise with the team. I'm not sure if any of these should be addressed as part of this PR, but wanted to go ahead and raise them her in case.

1. Should there be a knob in the Controls panel to change instance type (e.g. eval vs production/standard)? The message we display in the following example displays only to users with a subscription (we don't want to suggest that eval/trial users contact support). So the main difference is that the eval/trial users should not see this text: 
    > If your Kafka instance continues to take longer than expected, open a support case.
    
    This is noted in the original description that Cliodhna provided, but it would be good to find a way to show that in this component. I'm not sure if having that as a knob or extra story is preferred. 
    image

2. Should we include the values we get from the api in these examples? For example, the Metrics examples provide details about what kind of information we're getting from the api. For this, I guess it would be specifically what status we get from the api that results in each of the variations we show. Is that level of detail helpful to include in our documentation of our components?

ajaypratap003:
> Thanks for making the updates, @ajaypratap003. Everything looks good!
> 
> Reviewing these examples, I had a couple of questions that I wanted to raise with the team. I'm not sure if any of these should be addressed as part of this PR, but wanted to go ahead and raise them her in case.
> 
> 1. Should there be a knob in the Controls panel to change instance type (e.g. eval vs production/standard)? The message we display in the following example displays only to users with a subscription (we don't want to suggest that eval/trial users contact support). So the main difference is that the eval/trial users should not see this text:
>    > If your Kafka instance continues to take longer than expected, open a support case.
>    
>    
>    This is noted in the original description that Cliodhna provided, but it would be good to find a way to show that in this component. I'm not sure if having that as a knob or extra story is preferred.
>    image

Now there is no knob for Controls panel to change instance type (e.g. eval vs production/standard). I can add it if required.
> 2. Should we include the values we get from the api in these examples? For example, the Metrics examples provide details about what kind of information we're getting from the api. For this, I guess it would be specifically what status we get from the api that results in each of the variations we show. Is that level of detail helpful to include in our documentation of our components?

This components don't need more details as we had for Metrics. It needs only kafka status which we are getting from api.
Based on current kafka status component will change the view.
image

src/KafkaInstanceStatusPopover/KafkaInstanceStatusPopover.stories.tsx
riccardo-forina: I think you can, yes!

riccardo-forina (approved):
Thank you, there are things I'd like to address and extend but I'll do a follow up PR for that

PR #145, opened 2022-02-04
feat: KafkaInstanceStatus with popover and label

riccardo-forina:
I did some work on the Kafka status popover to prepare the field for https://github.com/redhat-developer/app-services-ui-components/pull/132

Changes

* the main exposed component is `KafkaInstanceStatus`. It's the composition of the popover and the label (a temporary one right now, it will be the outcome of the PR above).
* it takes the status of the instance and the creation date, which is used to automatically determine the alert to show to the user.
* It also re-exposes the status text component as `KafkaInstanceStatusLabel`, in case we need to show it somewhere without a popover.
* some refactoring of the code


ajaypratap003 (approved):

cfullam1 (commented):
LGTM! I think there was considerations of changing some of the statuses e.g. "Creation in progress" -> "Creating"

jgiardino (approved):
LGTM

riccardo-forina:
Updated to use your PoC for the status label @cfullam1 !

cfullam1 (approved):

PR #157, opened 2022-02-11
feat: add  acls treeview component

ajaypratap003:
https://issues.redhat.com/browse/MGDSTRM-6906

image


src/utils/types.ts
```diff
@@ -0,0 +1,20 @@
+enum AclOperation {
```
riccardo-forina, line 1: Please move these types to `Kafka/ManageKafkaPermissions`

src/utils/types.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+enum AclResourceType {
+  Group = "GROUP",
+  Topic = "TOPIC",
+  Cluster = "CLUSTER",
+  TransactionalId = "TRANSACTIONAL_ID",
+}
+
+export { AclOperation, AclResourceType };
```
riccardo-forina, line 20:
```suggestion
```

src/utils/types.ts
```diff
@@ -0,0 +1,20 @@
...
+  All = "ALL",
+  Read = "READ",
+  Write = "WRITE",
+  Create = "CREATE",
+  Delete = "DELETE",
+  Alter = "ALTER",
+  Describe = "DESCRIBE",
+  DescribeConfigs = "DESCRIBE_CONFIGS",
+  AlterConfigs = "ALTER_CONFIGS",
+}
```
riccardo-forina, line 11:
```suggestion
export type AclOperation =
  | "ALL"
  | "READ"
  | "WRITE"
  | "CREATE"
  | "DELETE"
  | "ALTER"
  | "DESCRIBE"
  | "DESCRIBE_CONFIGS"
  | "ALTER_CONFIGS";
```

`enum` is very badly supported both in TS and JS. A simpler type works better.

src/utils/types.ts
```diff
@@ -0,0 +1,20 @@
...
+
+enum AclResourceType {
+  Group = "GROUP",
+  Topic = "TOPIC",
+  Cluster = "CLUSTER",
+  TransactionalId = "TRANSACTIONAL_ID",
+}
```
riccardo-forina, line 18:
```suggestion

export type AclResourceType =
  | "GROUP"
  | "TOPIC"
  | "CLUSTER"
  | "TRANSACTIONAL_ID";
```

src/shared/SolidLabel/SolidLabel.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type SolidLabelProps = {
+  variant: AclResourceType;
+};
+
+type Variant =
+  | {
+      labelColor: LabelProps["color"];
+      content: string;
+    }
+  | undefined;
```
riccardo-forina, line 15:
```suggestion
```

Always favor stricter types. We don't need to support unknown groups, so we don't need to allow for unknown variants.

src/Kafka/AclsTreeView/AclsTreeView.tsx
```diff
@@ -0,0 +1,403 @@
...
+  setSelectedItems: (items: TreeViewDataItem[]) => void;
+};
+
+export const AclsTreeView: React.VFC<AclsTreeViewProps> = ({
+  setSelectedItems,
+}) => {
+  const { t } = useTranslation("manage-kafka-permissions");
+
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [checkedItems, setCheckedItems] = useState<TreeViewDataItem[]>([]);
```
riccardo-forina, line 27: I think we should be [lifting this state up](https://reactjs.org/docs/lifting-state-up.html). We will have a toolbar where this filter component will be displayed, and that toolbar will probably have a "Clear all filter" button to reset the filter to the initial state (which might be a previously saved one if that feature will get the ok).

riccardo-forina (changes requested):
Thanks Ajay, I left some comments and suggestions about how to organize this component.

Also please add tests after you are done.

src/shared/SolidLabel/SolidLabel.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
```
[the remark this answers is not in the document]
ajaypratap003: As we are moving complete `Manage Permissions`  code in ui-components repo. we need it in create and existing table component. That's why I added it here. This is common component. 
Let me know if you still want to move it from here.

src/utils/types.ts
ajaypratap003: We need value, not type checking. See below code where we need enum. i think we can't access value from type.
```
export const SolidLabel: React.VFC<SolidLabelProps> = ({ variant }) => {
  const getVariant = (): Variant => {
    switch (variant) {
      case AclResourceType.Group:
        return { labelColor: "green", content: "G" };
      case AclResourceType.Cluster:
        return { labelColor: "grey", content: "KI" };
      case AclResourceType.Topic:
        return { labelColor: "blue", content: "T" };
      case AclResourceType.TransactionalId:
        return { labelColor: "orange", content: "TI" };
      default:
        return;
    }
  };
  ```

Comment (riccardo-forina) [reply src/utils/types.ts:11]: Sure you can. That type is simply a `string` restricted to just those values.
In the switch you can use the expected string, Typescript will suggest the right ones and all will be typesafe. Eg: `case "GROUP"`

File: .eslintrc [hunk 8]
```diff
@@ -14,7 +14,7 @@
   ],
   "rules": {
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "error",
+    "react-hooks/exhaustive-deps": "off",
```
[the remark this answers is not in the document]
ajaypratap003: Actually, eslint forcing to enter everything as  dependencies which we used inside useEffect hook. we need to add dependencies on which we want to re-execute useEffect hook code. If we will add everything as dependency, it will cause the issue of infinite rendering.
See below error for reference:
```
  useEffect(() => {
    window.addEventListener("keydown", handleMenuKeys);
    window.addEventListener("click", handleClickOutside);
    return () => {
      window.removeEventListener("keydown", handleMenuKeys);
      window.removeEventListener("click", handleClickOutside);
    };
  }, [isOpen, menuRef]);
```

`
 320:6  error  React Hook useEffect has missing dependencies: 'handleClickOutside' and 'handleMenuKeys'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
`

I don't think so we really need to pass functions `handleClickOutside` `handleMenuKeys` as dependencies. 

Earlier we tried to fixed  this kind of all eslint issue by adding dependencies but it was causing issue of infinite re-rendering. 

Let me know the better solution. I think we can add rule as `warning` instead of `error`. wdyt ?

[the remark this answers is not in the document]
riccardo-forina: You have to wrap the missing dependencies with `useCallback`, and specify for them the right dependencies too. `handleMenuKeys` depends on `isOpen` for example.

ajaypratap003:
@riccardo-forina Thanks for review PR and adding suggestions. Most of the feedback I've implemented and added few inline comments for hook rule and eum. Please revisit PR and let me know if anything missing.

src/Kafka/ManageKafkaPermissions/components/ResourceTypeOperationFilter/ResourceTypeOperationFilter.tsx
```diff
@@ -0,0 +1,399 @@
...
+  MenuList,
+  Popper,
+  TreeView,
+  TreeViewDataItem,
+} from "@patternfly/react-core";
+import { SolidLabel } from "../SolidLabel";
+import { AclResourceType, AclOperation } from "../../types";
+
+export type ResourceTypeOperationFilterProps = {
+  setCheckedItems: (items: TreeViewDataItem[]) => void;
```
riccardo-forina, line 18:
```suggestion
  onCheckedItemsChange: (items: TreeViewDataItem[]) => void;
```

src/Kafka/ManageKafkaPermissions/components/ResourceTypeOperationFilter/ResourceTypeOperationFilter.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <ResourceTypeOperationFilter
+      setCheckedItems={setCheckedItems}
+      checkedItems={checkedItems}
+    />
+  );
+};
+
+export const ResourceTypeOperationFilterStory = Template.bind({});
+ResourceTypeOperationFilterStory.storyName = "ResourceTypeOperationFilter";
```
riccardo-forina, line 22: Now that you can pass the checkedItems, it would be nice to have a few stories where we show how the menu look like depending on the various selection options.

Also, test the stories with jest tests. Check [here](https://github.com/redhat-developer/app-services-ui-components/blob/main/src/shared/FormatDate/FormatDate.test.tsx) for an example about how to do that.

src/Kafka/ManageKafkaPermissions/components/ResourceTypeOperationFilter/ResourceTypeOperationFilter.tsx
```diff
@@ -350,7 +358,7 @@ export const ResourceTypeOperationFilter: React.VFC<
   const transactionalIdMapped = transactionalIdOptions.map(mapTree);
 
   const menu = (
-    <Menu ref={menuRef}>
+    <Menu ref={menuRef as React.RefObject<HTMLDivElement>}>
```
riccardo-forina, line 361: Try not to use `as` if you can. The proper way to specify the ref with Typescript would be:
`const toggleRef = useRef<HTMLButtonElement>(null);`

Note the `null`, it should fix the error.

src/Kafka/ManageKafkaPermissions/components/ResourceTypeOperationFilter/ResourceTypeOperationFilter.tsx
```diff
@@ -233,7 +233,7 @@ export const ResourceTypeOperationFilter: React.VFC<
   const filterItems = (
     item: TreeViewDataItem,
     checkedItem: TreeViewDataItem
-  ) => {
+  ): TreeViewDataItem | boolean | undefined => {
```
riccardo-forina, line 236: This type is a bit confusing, wouldn't `TreeViewDataItem | boolean` be enough, and return `false` instead of `undefined` be better?

riccardo-forina:
Overall it looks good, but it's still missing the stories I suggested and the Jest tests.

riccardo-forina:
The codecov error raises an interesting point, the functions you have in the render body of the `ResourceTypeOperationFilter` component look like they can be extracted to live outside the render and have unit tests for them too.

PR #164, opened 2022-02-16
feat: Remove Button Component for access management permission modal

HemaHG:
Remove Button Component for access management permission modal
part of https://issues.redhat.com/browse/MGDSTRM-7513 
 
 where to find stories 
Shared >  RemoveButton > RemoveButton


src/Kafka/ManageKafkaPermissions/components/RemoveButtonCell.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { t } = useTranslation();
+  return (
+    <div className="pf-u-display-flex pf-u-justify-content-flex-end">
+      <Tooltip
+        content={t(
+          "kafka:manage_permissions_dialog.assign_permissions.remove_row_help"
+        )}
+      >
+        <Button variant="link" icon={<TrashIcon />} onClick={() => removeRow} />
+      </Tooltip>
```
riccardo-forina, line 22: I wonder if this shouldn't be a generic component under the Shared folder. @jgiardino what do you think? I'm talking about the button with the tooltip only, not the cell thingy used to align it to the right.

jgiardino: I can't think of other places where we use or might use an icon-only button, but I still think this is valid as a generic component, since it's a common pattern to have an icon button with tooltip text. 

As a more generic button, would that require the ability to specify the icon and the tooltip text? 

On a separate topic, should the `variant="link"` be used here? When I look at the [PF example](https://www.patternfly.org/v4/components/button), they use `variant="plain"` for the icon-only button. Also looking at one of our [design documents](https://marvelapp.com/prototype/2cb178b6/screen/83633198) for the manage permissions modal, the design seems to be consistent with the PF example.

riccardo-forina: I'd avoid making it too generic, that's the domain of PatternFly's components. We should stick to specialized - albeit small - components that don't offer too much customization.

+1 about the variant, good catch!

jgiardino (approved):
LGTM!

PR #175, opened 2022-02-22
chore: Add account selection of manage access modal to storybooks

suyash-naithani:
This PR includes:

- Adding a modal for manage access
- Select component for selecting an account which is step 1 of the manage access modal 
- Improvements on managing validation of the selected account


suyash-naithani:
@riccardo-forina I created an interactive example which works as expected, but on other stories, in SelectAccount.tsx, the onSelect function does not seem to be saving the value correctly and logging it into the actions tab in storybook. What I see in actions tab is just the name of the function. Let me know if you can figure out why the story is not working as expected  barring the interactive story

jgiardino:
Hi @suyash-naithani, thanks for working on these! The examples look good, here is just one thing I noticed that could be improved from a UX perspective...

In the **No service or user accounts** example, there should always be an option for **All accounts** in the menu—the lack of accounts shouldn't prevent the user from creating an ACL for all accounts. 

I do like that there is a statement in the menu "No results found", but is it possible to keep all the other elements of the menu present (i.e. "All accounts" option, section headers, and dividers)? For example, instead of what you see in the screencap below, the menu would have these items:
```
All accounts
Uses a wildcard (*) to assign permissions to all user and service accounts
————————
Service accounts
No results found
————————
User accounts
No results found
```
This way, we have a proper empty state for each section.

image

src/Kafka/ManageKafkaPermissions/components/ManagePermissionsModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <Modal
+      id="manage-permissions-modal"
+      variant={"large"}
+      isOpen={true}
+      aria-label={t("manage_permissions_dialog.aria_label")}
+      position="top"
+      title={t("manage_permissions_dialog.title")}
+      showClose={true}
+      aria-describedby="modal-message"
```
jgiardino, line 81: The value defined here should reference an `id` that's defined for the primary message of this modal. We don't really have a primary message, but we do state the name of the kafka instance. To make this functional, we should add `id="modal-message"` to the `<FormGroup>` earlier in this document that includes the kafka instance name. This way, when a screen reader announces the modal, it will also provide the kafka instance name for additional context to the user. 

```
      <FormGroup
        fieldId="kafka-instance-name"
        label={t("manage_permissions_dialog.kafka_instance_title")}
        id="modal-message"
      >
        {kafkaName}
      </FormGroup>
```

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,187 @@
...
+      );
+  };
+
+  return (
+    <FormGroupWithPopover
+      labelHead={t("manage_permissions_dialog.account_id_title")}
+      fieldId="kafka-instance-name"
+      fieldLabel={t("manage_permissions_dialog.account_id_title")}
+      labelBody={t("manage_permissions_dialog.account_id_help")}
+      buttonAriaLabel={t("manage_permissions_dialog.account_id_aria")}
```
jgiardino, line 153: I know this text string isn't defined in this PR this is defined, but when I look at the value being used for the Help button aria-label, it just says "The account"

If we follow the example in PF, then a better label for this button would be "More info for Account field"

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,187 @@
...
+        onClear={clearSelection}
+        selections={id.value}
+        onFilter={customFilter}
+        isOpen={isOpen}
+        placeholderText={t(
+          "manage_permissions_dialog.account_id_typeahead_placeholder"
+        )}
+        placeholder={t(
+          "manage_permissions_dialog.account_id_typeahead_placeholder"
+        )}
```
jgiardino, line 176: When I look at the [PF example for typeahead select](https://www.patternfly.org/v4/components/select#typeahead), I see a `placeholderText` prop, but not a `placeholder` prop. This is what provides the "Select an account" that displays inside the typeahead input. It would be good to check if we even need the `placeholder` here and remove it if not needed.

src/Kafka/ManageKafkaPermissions/components/ManagePermissionsModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+export enum PrincipalType {
+  UserAccount = "USER_ACCOUNT",
+  ServiceAccount = "SERVICE_ACCOUNT",
+}
+
+export type Principal = {
+  id: string;
+  principalType?: PrincipalType;
+  displayName?: string;
+};
```
riccardo-forina, line 22: These types should live in a `types.ts` file, probably a level above the components folder.

I'm not sure about the use of the word `Principal` here. In the UI we show this as an `Account`, we should stick to that name.

```suggestion
export enum AccountType {
  User = "USER_ACCOUNT",
  Service = "SERVICE_ACCOUNT",
}

export type Account = {
  id: string;
  type: AccountType;
  displayName?: string;
};
```

I also commented in the past in some other PR that [enums aren't a great choice](https://blog.logrocket.com/why-typescript-enums-suck/) and I stand my point, although it's not a hill I'd die on.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,187 @@
...
+          </SelectOption>
+        </SelectGroup>,
+        <Divider key="all_accounts_divider" />,
+        <SelectGroup
+          label={t(
+            "manage_permissions_dialog.all_accounts_service_account_group"
+          )}
+          key="service_accounts_group"
+        >
+          {initialOptions
```
riccardo-forina, line 84: It's a good practice to keep the logic on the jsx at a minimum to make it easier to read. This filtering and sorting should be done before the render, and the jsx moved inside the return of the render.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,187 @@
...
+                {principal.id}
+              </SelectOption>
+            ))}
+        </SelectGroup>,
+        <Divider key="user_account_divider" />,
+        <SelectGroup
+          label={t("manage_permissions_dialog.all_accounts_user_account_group")}
+          key="user_accounts_group"
+        >
+          {initialOptions
```
riccardo-forina, line 109: As above.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,187 @@
...
+      labelBody={t("manage_permissions_dialog.account_id_help")}
+      buttonAriaLabel={t("manage_permissions_dialog.account_id_aria")}
+      isRequired={true}
+      helperTextInvalid={id.errorMessage}
+      validated={id.validated || ValidatedOptions.default}
+    >
+      <Select
+        data-testid="acls-select-account"
+        variant={SelectVariant.typeahead}
+        className="kafka-ui--select--limit-height"
```
riccardo-forina, line 161: A CSS class without a companion rule? Please remove this if we don't need it, and if we do there is a [naming convention that needs to be followed](https://redhat-developer.github.io/app-services-ui-components/iframe.html?id=contributing--page&args=&viewMode=story#classnames-and-id-naming-conventions) when naming classes

riccardo-forina (changes requested):
Thanks Suyash, I left a few comments.

suyash-naithani:
@jgiardino I have implemented your suggestions in the stories. There is one more thing that I would like to discuss regarding the No Service or User Accounts story. You mentioned that we would like to have the option for selecting all accounts available in the select option. Would we want that option to be selectable, and if so, on selecting the particular option, it is accepted as a valid input and the user is able to click on next, even though he had no accounts to select from. 
Maybe not having the option for selecting all accounts when there are no accounts would be a better approach, or make the all accounts option non clickable if accounts length is zero.

suyash-naithani:
@riccardo-forina  Thanks for the review. I have tried to implement your review comments, let me know what you think :)

jgiardino:
Hi @suyash-naithani, thanks for making updates!

1. In response to your question above:
    > I have implemented your suggestions in the stories. There is one more thing that I would like to discuss regarding the No Service or User Accounts story. You mentioned that we would like to have the option for selecting all accounts available in the select option. Would we want that option to be selectable, and if so, on selecting the particular option, it is accepted as a valid input and the user is able to click on next, even though he had no accounts to select from.
Maybe not having the option for selecting all accounts when there are no accounts would be a better approach, or make the all accounts option non clickable if accounts length is zero.
    - If the API supports this and there's no technical reason that prevents the user from assigning a permission to "all accounts", then we shouldn't block the user from doing this via the UI. The "All accounts" option is really more of a wildcard value. We aren't creating ACLs for all the accounts that exist in the org, but instead create an ACL with a wildcard value `*` specified instead of the account ID of a specific account. 
2. I'm not seeing the PF styles being pulled into the Select Account stories for the form label, help button (?) or required asterisk (*). I do see styles display as expected for this component inside the Managed Kafka Permission story:

    | Select Account | Managed Kafka Permission |
    | -- | -- |
    | styles aren't being applied to the form label and adjacent UI elements | styles are being applied to the form label and adjacent UI elements | 
    | image | image |

3. A nice future enhancement to consider is updating the account ID and display name values to have more realistic values. This would help with testing that filtering works as expected, and give other team members a better idea of how values would display. I don't see this as critical for this PR, though.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,206 @@
...
+  };
+  const options = [
+    <SelectGroup key="all_accounts_group">
+      <SelectOption
+        key="*"
+        value="*"
+        description={t("manage_permissions_dialog.all_accounts_description")}
+      >
+        {t("manage_permissions_dialog.all_accounts_title")}
+      </SelectOption>
```
jgiardino, line 125: I don't think the "All accounts" option should be inside a `<SelectGroup>` component. Since this section of the menu doesn't have a group label, including `<SelectGroup>` without a `label` prop results in extra space above this option, making it look inconsistent with other menus. 

image

ajaypratap003:
> Hi @suyash-naithani, thanks for making updates!
> 
> 1. In response to your question above:
>    > I have implemented your suggestions in the stories. There is one more thing that I would like to discuss regarding the No Service or User Accounts story. You mentioned that we would like to have the option for selecting all accounts available in the select option. Would we want that option to be selectable, and if so, on selecting the particular option, it is accepted as a valid input and the user is able to click on next, even though he had no accounts to select from.
>    > Maybe not having the option for selecting all accounts when there are no accounts would be a better approach, or make the all accounts option non clickable if accounts length is zero.
>    
>    
>    
>    * If the API supports this and there's no technical reason that prevents the user from assigning a permission to "all accounts", then we shouldn't block the user from doing this via the UI. The "All accounts" option is really more of a wildcard value. We aren't creating ACLs for all the accounts that exist in the org, but instead create an ACL with a wildcard value `*` specified instead of the account ID of a specific account.
> 2. I'm not seeing the PF styles being pulled into the Select Account stories for the form label, help button (?) or required asterisk (*). I do see styles display as expected for this component inside the Managed Kafka Permission story:
>    
>    
>    
>    Select Account
>    Managed Kafka Permission
>    
>    
>    
>    
>    styles aren't being applied to the form label and adjacent UI elements
>    styles are being applied to the form label and adjacent UI elements
>    
>    
>    image
>    image
> 3. A nice future enhancement to consider is updating the account ID and display name values to have more realistic values. This would help with testing that filtering works as expected, and give other team members a better idea of how values would display. I don't see this as critical for this PR, though.

@jgiardino  @suyash-naithani 
PF styles will not work with individual `SelectAccount` label which is using FormGroup and `FormGroup` works with `From`.
There are two ways to handle it if you want to sse it in storybook individually. 
1. Add `Form` tag in storybook component
2. Remove `FormGroup` from SelectAcoount  component and wrap it in parent component i..e Modal

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+  id,
+  accounts,
+  onEscapeModal,
+  validated,
+  onChangeValidation,
+}) => {
+  const { t } = useTranslation(["manage-kafka-permissions"]);
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  const onToggle = (newState: boolean | ((prevState: boolean) => boolean)) => {
```
ajaypratap003, line 43: I'm curious, what is the use of second argument `prevState` here ?

ajaypratap003 (approved):

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+    _: React.ChangeEvent<HTMLInputElement> | null,
+    value: string
+  ) => {
+    if (!value) {
+      return options;
+    }
+
+    const input = new RegExp(value, "i");
+    return options
+      .filter((accounts) => Array.isArray(accounts.props.children))
```
riccardo-forina, line 154: `children` should be manipulated using the [`React.Children` functions](https://reactjs.org/docs/react-api.html#reactchildren)

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+  value: T;
+  validated?: ValidatedOptions;
+  errorMessage?: string;
+};
+
+export type SelectAccountProps = {
+  id: string | undefined | SelectOptionObject;
+  onChangeAccount: React.Dispatch<
+    React.SetStateAction<string | undefined | SelectOptionObject>
+  >;
```
riccardo-forina, line 25:
```suggestion
  onChangeAccount: (value: string) => void;
```

I commented on this earlier, you should not force the consumer of the component into using React's state. Just make the new value available and let the parent component deal with it however it wants.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+  accounts: Account[];
+  onEscapeModal: (closes: boolean) => void;
+  validated: ValidatedOptions;
+  onChangeValidation: (value: ValidatedOptions) => void;
```
riccardo-forina, line 29: Not sure why the UI component should tell the parent if the selected value is valid or not. It should be the parent that knows if the value is valid, and the validation state is indeed being passed.

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+} from "@patternfly/react-core";
+
+export type Validated<T> = {
+  value: T;
+  validated?: ValidatedOptions;
+  errorMessage?: string;
+};
+
+export type SelectAccountProps = {
+  id: string | undefined | SelectOptionObject;
```
riccardo-forina, line 22:
```suggestion
  value: string | undefined
```

Naming is hard, I know. But `id` in HTML has its own meaning. A common way to provide the value of a field is via a prop called `value`. I'd use that. 

Also, why should the parent deal with the complexities of the PF's Select component? The idea of writing specialized components is to make it easier to work with the problem at hand. If an account is a string, we should talk about strings. Not the weird `string | undefined | SelectOptionObject` that comes from the PF's Select

src/Kafka/ManageKafkaPermissions/ManageKafkaPermissions.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useRef, useState } from "react";
+import { useTranslation } from "react-i18next";
+import { FormGroupWithPopover } from "../..";
+import { SubmitButton } from "./components/PermissionsModalSubmitButton";
+import { SelectAccount } from "./components/SelectAccount";
+import { Account } from "./types";
+
+type ManageKafkaPermissionsProps = {
+  accounts: Account[];
+  hideModal: () => void;
```
riccardo-forina, line 18:
```suggestion
  onCancel: () => void;
```

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
```diff
@@ -0,0 +1,194 @@
...
+          ))
+      : [
+          <SelectOption isNoResultsOption={true} isDisabled={true} key={1}>
+            {t("manage_permissions_dialog.no_results_found")}
+          </SelectOption>,
+        ];
+  };
+
+  const onSelect: SelectProps["onSelect"] = (_, selection) => {
+    onChangeAccount(selection);
```
suyash-naithani, line 113: @riccardo-forina  The reason I added SelectOptionObject type instead of just going with a string|undefined type for the state of value is the type error I'm getting here. 
**_Argument of type 'string | SelectOptionObject' is not assignable to parameter of type 'string | undefined'.
  Type 'SelectOptionObject' is not assignable to type 'string'_**
Any suggestions on how I can remove that?

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
suyash-naithani: I'll fix the naming :) . I had a question about the type, I've posted it [here](https://github.com/redhat-developer/app-services-ui-components/pull/175#discussion_r820959213)

src/Kafka/ManageKafkaPermissions/components/SelectAccount.tsx
riccardo-forina: In this case it's ok to do a cast `as string` since you are feeding strings to the select.

riccardo-forina (approved):

PR #176, opened 2022-02-22
feat: add review permissions table and create storybook

ajaypratap003:


src/Kafka/ManageKafkaPermissions/components/ReviewPermissionsTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+          "manage_permissions_dialog.assign_permissions.remove_row_help"
+        )}
+      >
+        <Button
+          variant="link"
+          icon={<TrashIcon />}
+          onClick={() => onRemoveAcl(acl)}
+        />
+      </Tooltip>
+    );
```
jgiardino, line 54: @HemaHG has a PR open for this exact component: https://github.com/redhat-developer/app-services-ui-components/pull/164

Should we merge her PR and then use that component here instead?

src/Kafka/ManageKafkaPermissions/components/Cells.tsx
```diff
@@ -0,0 +1,74 @@
...
+      return t("cells.pattern_type_prefixed");
+    } else {
+      return t("cells.pattern_type_literal");
+    }
+  };
+  if (item.resourceType === AclResourceType.Cluster) {
+    return {
+      title: (
+        <div>
+          <SolidLabel variant={item.resourceType} />{" "}
```
jgiardino, line 34: This may be a question for @riccardo-forina when he's back, but should this component be surfaced as a story in storybook?

jgiardino:
These updates look great, @ajaypratap003! There are a couple of things I noticed while reviewing them:

1. Can you add an example of a `Deny` permission so that we have an example that illustrates how the label should display? You can see an example of a permission with `Deny` here: https://marvelapp.com/prototype/hc64c9a/screen/84139693
2. We have designs for the empty state of the table. Could you update the empty table example to display the empty state instead? In this case the table is just replaced with the text "No existing permissions".
    * The final design was provided here _but_ this also shows the empty state inside an expandable section, which we aren't using yet: https://marvelapp.com/prototype/hc64c9a/screen/84198391
    * This is a previous design where the text wasn't final yet, but you can see how the empty state displays with the current section header: https://marvelapp.com/prototype/2ci9id46/screen/84082998
3. I don't see any tests created for these examples yet.

src/Kafka/ManageKafkaPermissions/components/ReviewPermissionsTable.tsx
HemaHG: yes I create the remove component to use in the permission table

src/Kafka/ManageKafkaPermissions/components/ReviewPermissionsTable.tsx
```diff
@@ -0,0 +1,129 @@
...
+    },
+    {
+      title: t("table.permissions_column_title"),
+      columnTransforms: [cellWidth(20)],
+    },
+    {
+      title: "",
+      columnTransforms: [cellWidth(20)],
+    },
+  ] as ICell[];
```
riccardo-forina, line 105: Please, use the TableComposable component for tables https://www.patternfly.org/v4/components/table

The old syntax is really bad with typescript and hard to work with.

src/Kafka/ManageKafkaPermissions/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { sentenceCase } from "sentence-case";
+import { AclResourceType } from "./types";
+
+export const displayName = (resourceType: AclResourceType): string => {
```
riccardo-forina, line 4: This should be a react component, and the hardcoded strings should be under translations

riccardo-forina (commented):
Thank you Ajay, I quickly checked this from my phone and left a few comments.

jgiardino:
Thanks for making the requested updates! The only issue I see is with the following story:

image

The issues I noticed are:

1. This text should include a period (.) at the end.
2. This text doesn't use the expected styles. It should use the same styles as other text that displays below the section headers. You can see an example with our current implementation, as shown below with the text "Review the list of existing permissions..."

image

riccardo-forina:
I have done some cleanup and refactoring. 

The "No existing permissions story" has been removed, this component deals only with existing permissions. If there are no permissions the parent component will handle the case with the appropriate text where it needs to be placed.

The idea behind the refactoring is to lift all required logic to the main table component.

Stories have been simplified to do less and rely more on static data. To do so I changed the types slightly (no more enums, it makes string-based types too complex to work with).

I also renamed the `SolidLabel` component to `ResourceTypeComponent`. It's better to describe what a component does, rather than how it does it.

ajaypratap003:
> I have done some cleanup and refactoring.
> 
> The "No existing permissions story" has been removed, this component deals only with existing permissions. If there are no permissions the parent component will handle the case with the appropriate text where it needs to be placed.
> 
> The idea behind the refactoring is to lift all required logic to the main table component.
> 
> Stories have been simplified to do less and rely more on static data. To do so I changed the types slightly (no more enums, it makes string-based types too complex to work with).
> 
> I also renamed the `SolidLabel` component to `ResourceTypeComponent`. It's better to describe what a component does, rather than how it does it.

@riccardo-forina Thanks for the pushing changes. can we merge this PR ?

riccardo-forina (approved):

PR #190, opened 2022-03-02
chore: adding plain inline alert to modal

cfullam1:


src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
```diff
@@ -286,6 +288,18 @@ export const CreateKafkaInstance: FunctionComponent<
...
       </Flex>
+      <Flex>
+        <FlexItem>
+          <Alert
+            className="mas-m-modalTop"
+            customIcon={<OutlinedClockIcon />}
+            variant="info"
+            isInline
+            isPlain
+            title="Your Kafka instance will be ready for use shortly after creation."
```
riccardo-forina, line 299: This string should go in locales/en/create-kafka-instance.json

riccardo-forina (approved):
Thanks, it looks good. I'll approve but please move the string I commented on to the translation file before merging

src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
cfullam1: Ah thank you @riccardo-forina, just added that in

PR #191, opened 2022-03-03
feat: Consumer groups Drawer

HemaHG:
Part of https://issues.redhat.com/browse/MGDSTRM-7703

move the code related to the table inside the drawer in the consumer groups 
Screenshot from 2022-03-03 19-45-46

where to find stories
kafka > consumer groups > components > ConsumerGroupDetails

Consumer group at kafka level
kafkalevel

Consumer group at topic level
ConsumerGroupAttopicLevel


src/Kafka/ConsumerGroups/components/ConsumerGroupDetail.tsx
```diff
@@ -0,0 +1,197 @@
...
+            <Text component={TextVariants.h4}>
+              {consumerGroupByTopic
+                ? t("consumerGroup.partitions_with_lag_for_topic")
+                : t("consumerGroup.partitions_with_lag")}{" "}
+              <ConsumerGroupPopover
+                title={t("consumerGroup.partitions_with_lag_name")}
+                description={t("consumerGroup.partitions_with_lag_description")}
+              />
+            </Text>
+            <div>
```
HemaHG, line 71: @jgiardino , @riccardo-forina earlier instead of <div>,  <Text component={TextVariants.p}> was used but the test was failing with the below error so Changed to div, please suggest me if it's a good approach

riccardo-forina: I think you can remove the wrapper completely. The error is right, you can't have an h2 inside a p, but we just have an h2 so I don't see the need to wrap it into a div

src/Kafka/ConsumerGroups/components/ConsumerGroupDetail.tsx
```diff
@@ -0,0 +1,198 @@
...
+                  </Th>,
+                ]}
+          </Tr>
+        </Thead>
+        <Tbody>
+          {(consumerGroupDetail &&
+            consumerGroupDetail.consumers.map((consumergroup) => {
+              return (
+                <Tr key={consumergroup.groupId}>
+                  {consumerGroupByTopic
```
riccardo-forina, line 142: There are 3 of these ifs in the code. It makes the code difficult to follow. 

I'd create two different components and have the parent component use one or another depending on the need.

src/Kafka/ConsumerGroups/components/ConsumerGroupDetails/ConsumerGroupDetail.tsx
```diff
@@ -0,0 +1,23 @@
+import { FunctionComponent } from "react";
```
riccardo-forina, line 1: Since this is the feature component it should live in the parent folder and have the same name as the finder, so `src/Kafka/ConsumerGroups/ConsumerGroup.tsx`. The exported component should also have the same name as the file.

Also, please flatten up the components folder, we don't need a folder per component.

This is described [in the docs](https://redhat-developer.github.io/app-services-ui-components/?path=/story/folder-structure-convention--page#how-to-structure-a-component)

src/Kafka/ConsumerGroups/components/ConsumerGroupDetails/ConsumerGroupByKafka.tsx
```diff
@@ -0,0 +1,135 @@
...
+  Tr,
+} from "@patternfly/react-table";
+import { FunctionComponent } from "react";
+import { useTranslation } from "react-i18next";
+import { ConsumerGroup } from "../../types";
+import { displayConsumerGroupState } from "../../utils";
+import { ConsumerGroupPopover } from "../ConsumerGroupPopover";
+
+export type ConsumerGroupByKafka = {
+  consumerGroupDetail: ConsumerGroup;
```
riccardo-forina, line 25: I understand it's easy to copy-paste the type from the sdk and use it here, but it's also a way to get around of the rule of no sdk dependency. 

Instead of an object here, I'd have specified many props, one for each visual element that needs to be displayed.

src/Kafka/ConsumerGroups/components/ConsumerGroupDetails/ConsumerGroupDetail.tsx
HemaHG: ConsumerGroupDetails is not the main feature, it's the internal component used to build the feature so I kept that in components folder

src/Kafka/ConsumerGroups/utils/utils.ts
```diff
@@ -0,0 +1,8 @@
+export enum displayConsumerGroupState {
+  STABLE = "Stable",
+  DEAD = "Dead",
+  EMPTY = "Empty",
+  COMPLETING_REBALANCE = "Completing rebalance",
+  PREPARING_REBALANCE = "Preparing rebalance",
+  UNKNOWN = "Unknown",
+}
```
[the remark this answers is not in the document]
HemaHG: I understood this point, but I am still not getting how to convert this enum to accept the i18n translation.
still working on this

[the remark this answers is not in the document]
riccardo-forina: Something like this minus all the jsx https://github.com/redhat-developer/app-services-ui-components/blob/main/src/Kafka/KafkaInstanceStatus/components/StatusLabel/StatusLabel.tsx

HemaHG:
Thanks for the review @riccardo-forina. I have addressed the review comments

riccardo-forina (approved):
Nice, thank you!

PR #193, opened 2022-03-04
chore: WIP Instance Actions Menu Dropdown

cfullam1:
https://issues.redhat.com/browse/MGDSTRM-4695


jgiardino:
Thanks for starting this! 

I shared this [spreadsheet](https://docs.google.com/spreadsheets/d/1P758NpEzwdrScecbpbgzAL1h3ZBdxRf0PyndiVB2-4Y/edit#gid=338479352) with you off-thread, but want to share it again here for easy reference. This includes an overview of all the actions we knew about months back when Abi was working on guidelines for how to organize items. 

Based on the options you created, the first option most closely matches the labels that Abi had defined in that document.

cfullam1:
Thank you @jgiardino ! I will remove the other stories and update the correct one. I was wondering what text to put in for the [primary data plane tab] under the "View informations" section?

PR #194, opened 2022-03-07
fix: add reached max limit for supported regions

ajaypratap003:
Implement feedback missed  on PR https://github.com/bf2fc6cc711aee1a0c2a/kas-ui/pull/1072/files#r817060682
Add below stories 
https://619e1921b3e38b003afeb2be-ryflrmybhy.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--load-supported-regions-and-disabled-unsupported

https://619e1921b3e38b003afeb2be-ryflrmybhy.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-trial-instance-for-supported-regions

image

### DIFF
--- a/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
+++ b/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
@@ -83,6 +83,7 @@ export const CreateKafkaInstance: FunctionComponent<
     regions,
     azOptions,
     availableProviders,
+    instanceAvailability,
 
     isNameTaken,
     isNameInvalid,
@@ -110,11 +111,6 @@ export const CreateKafkaInstance: FunctionComponent<
     onCreate,
   });
 
-  let { instanceAvailability } = useCreateKafkaInstanceMachine({
-    getAvailableProvidersAndDefaults,
-    onCreate,
-  });
-
   const onSubmit = useCallback(
     (event: FormEvent<HTMLFormElement>) => {
       event.preventDefault();
@@ -123,11 +119,7 @@ export const CreateKafkaInstance: FunctionComponent<
     [create]
   );
 
-  const isRegionsAvailable =
-    regions && regions?.some(({ isDisabled }) => isDisabled !== true);
-
-  const disableControls =
-    isLoading || isSaving || !canCreate || !isRegionsAvailable;
+  const disableControls = isLoading || isSaving || !canCreate;
 
   const nameValidation = isNameError ? "error" : "default";
   const providerValidation = isProviderError ? "error" : "default";
@@ -135,8 +127,6 @@ export const CreateKafkaInstance: FunctionComponent<
   const azValidation = isAzError ? "error" : "default";
   const disableAZTooltip =
     azOptions === undefined || (azOptions?.multi === true && azOptions.single);
-
-  if (!isRegionsAvailable) instanceAvailability = "trial-unavailable";
 
   return (
     <Modal
@@ -154,7 +144,7 @@ export const CreateKafkaInstance: FunctionComponent<
           variant="primary"
           type="submit"
           form={FORM_ID}
-          isDisabled={!canSave || !isRegionsAvailable}
+          isDisabled={!canSave}
           spinnerAriaValueText={t("common:submitting_request")}
           isLoading={isSaving}
           data-testid="modalCreateKafka-buttonSubmit"
@@ -231,7 +221,7 @@ export const CreateKafkaInstance: FunctionComponent<
                 value={region}
                 regions={regions || []}
                 onChange={setRegion}
-                isDisabled={disableControls || !isRegionsAvailable}
+                isDisabled={disableControls}
                 validated={regionValidation}
               />
             </FormGroup>
--- a/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
+++ b/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
@@ -83,7 +83,6 @@ export const CreateKafkaInstance: FunctionComponent<
     regions,
     azOptions,
     availableProviders,
-    instanceAvailability,
 
     isNameTaken,
     isNameInvalid,
@@ -111,6 +110,11 @@ export const CreateKafkaInstance: FunctionComponent<
     onCreate,
   });
 
+  let { instanceAvailability } = useCreateKafkaInstanceMachine({
+    getAvailableProvidersAndDefaults,
+    onCreate,
+  });
+
   const onSubmit = useCallback(
     (event: FormEvent<HTMLFormElement>) => {
       event.preventDefault();
@@ -119,7 +123,11 @@ export const CreateKafkaInstance: FunctionComponent<
     [create]
   );
 
-  const disableControls = isLoading || isSaving || !canCreate;
+  const isRegionsAvailable =
+    regions && regions?.some(({ isDisabled }) => isDisabled !== true);
+
+  const disableControls =
+    isLoading || isSaving || !canCreate || !isRegionsAvailable;
 
   const nameValidation = isNameError ? "error" : "default";
   const providerValidation = isProviderError ? "error" : "default";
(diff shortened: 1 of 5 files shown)


src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
```diff
@@ -128,6 +136,8 @@ export const CreateKafkaInstance: FunctionComponent<
   const disableAZTooltip =
     azOptions === undefined || (azOptions?.multi === true && azOptions.single);
 
+  if (!isRegionsAvailable) instanceAvailability = "trial-unavailable";
```
riccardo-forina, line 139: I'm ok with the changes but please move them into the hook. The idea is to have all the business logic in the hook and the UI as a simple consumer of that. This way you don't have to call the hook twice just to make a const a let

ajaypratap003: done

riccardo-forina:
Thank you. Actually thinking about this more, the changes should be done on the state machine, which is the actual owner of the business logic. Do you want to try doing that?

ajaypratap003:
> Thank you. Actually thinking about this more, the changes should be done on the state machine, which is the actual owner of the business logic. Do you want to try doing that?

Yes, I'll try. Do you mind to explain bit more, what changes exactly you want ?

ajaypratap003:
> @ajaypratap003 I'm having a hard time understanding what updates are included in this PR. Can you provide the names of the new or updated stories that need review?

Below are the stories
https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-supported-region-after-submit-create-form

https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-trial-instance-for-supported-regions

https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--load-supported-regions-and-disabled-unsupported

jgiardino:
Thanks for the clarification, @ajaypratap003. 

The first two links you share above appear to be duplicates. For each use case, I listed below the stories that you added and the stories that already exist.  I tried to clarify on Friday in our team chat that these stories already exist, but I think my message may have been overlooked. 

Can you take a look and identify how these existing stories are different from the new stories you added, and merge any changes that you see are needed? Visually they appear to be identical, but maybe there are some optimizations you made for being able to consume them 🤷‍♀️ 

My suggestion would be to only make changes that are absolutely necessary to support addressing the bug at hand (i.e. we want to display a more specific message instead of "Something went wrong").  Any further enhancements we make should be made as part of the next round of updates related to the larger Kafka instance sizes. 

| use case | New story from this PR | Existing story before this PR |
| -- | -- | -- |
| Selected region is no longer available on form submit | [Reached Max Limit For Supported Region After Submit Create Form](https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-supported-region-after-submit-create-form) | [Trial Unavailable On Form Submit](https://619e1921b3e38b003afeb2be-guvpmhdncg.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--trial-unavailable-on-form-submit) |
| No regions are available on form load | [Reached Max Limit For Trial Instance For Supported Regions](https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-trial-instance-for-supported-regions) | [Trial Unavailable on Form Load](https://619e1921b3e38b003afeb2be-guvpmhdncg.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--trial-unavailable-on-form-load) |

⚠️ One additional comment about the **Trial Unavailable On Form Submit** example — I know that in recent discussion it came up that it's possible these cases we planned for the eval/trial user where a region is out of capacity could happen for the user with a subscription, but it's much more rare for that case. The messages we created for the eval/trial user still work for a paying user. However, in this story, the blue inline alert about 48 hours is included. When we add the red "Kafka instance can't be created" to the kas-ui, is it possible to not show the blue "48 hours" message for users with a subscription, or do we need an explicit story to support that implementation in kas UI?

image

riccardo-forina:
> > Thank you. Actually thinking about this more, the changes should be done on the state machine, which is the actual owner of the business logic. Do you want to try doing that?
> 
> Yes, I'll try. Do you mind to explain bit more, what changes exactly you want ?

The way the component works, it's the parent that tells the component about the availability through the `getAvailableProvidersAndDefaults` function. I think it should be the parent the one that checks for the presence of at least one enabled region, and if there are none then tells the UI that there is no availability.

The changes here should be only about the possibility to have disabled regions.

ajaypratap003:
> Thanks for the clarification, @ajaypratap003.
> 
> The first two links you share above appear to be duplicates. For each use case, I listed below the stories that you added and the stories that already exist. I tried to clarify on Friday in our team chat that these stories already exist, but I think my message may have been overlooked.
> 
> Can you take a look and identify how these existing stories are different from the new stories you added, and merge any changes that you see are needed? Visually they appear to be identical, but maybe there are some optimizations you made for being able to consume them woman_shrugging
> 
> My suggestion would be to only make changes that are absolutely necessary to support addressing the bug at hand (i.e. we want to display a more specific message instead of "Something went wrong"). Any further enhancements we make should be made as part of the next round of updates related to the larger Kafka instance sizes.
> 
> use case	New story from this PR	Existing story before this PR
> Selected region is no longer available on form submit	[Reached Max Limit For Supported Region After Submit Create Form](https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-supported-region-after-submit-create-form)	[Trial Unavailable On Form Submit](https://619e1921b3e38b003afeb2be-guvpmhdncg.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--trial-unavailable-on-form-submit)
> No regions are available on form load	[Reached Max Limit For Trial Instance For Supported Regions](https://619e1921b3e38b003afeb2be-gjxdftdogi.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--reached-max-limit-for-trial-instance-for-supported-regions)	[Trial Unavailable on Form Load](https://619e1921b3e38b003afeb2be-guvpmhdncg.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--trial-unavailable-on-form-load)


@jgiardino 
First one seems same. I'll remove that. 
second one visually same but logically different. 
I'm showing warning message based on regions unavailability i.e. regions=[]
and second one based on instanceAvailability i.e. "trial-unavailable". Logically both edge cases are different but visually same.

> warning One additional comment about the **Trial Unavailable On Form Submit** example — I know that in recent discussion it came up that it's possible these cases we planned for the eval/trial user where a region is out of capacity could happen for the user with a subscription, but it's much more rare for that case. The messages we created for the eval/trial user still work for a paying user. However, in this story, the blue inline alert about 48 hours is included. When we add the red "Kafka instance can't be created" to the kas-ui, is it possible to not show the blue "48 hours" message for users with a subscription, or do we need an explicit story to support that implementation in kas UI?
> 

Earlier we had single message on modal but now we have two component one for Form level message and second for Modal level message. As your concern about 48 hours message that will be display only for trial/eval user, not for standard user. In this example we have trial instance that's why blue message display. But this will not be the case for standard instance.
Please let me know if I'm missing something here.
> image

ajaypratap003:
> > > Thank you. Actually thinking about this more, the changes should be done on the state machine, which is the actual owner of the business logic. Do you want to try doing that?
> > 
> > 
> > Yes, I'll try. Do you mind to explain bit more, what changes exactly you want ?
> 
> The way the component works, it's the parent that tells the component about the availability through the `getAvailableProvidersAndDefaults` function. I think it should be the parent the one that checks for the presence of at least one enabled region, and if there are none then tells the UI that there is no availability.
> 
> The changes here should be only about the possibility to have disabled regions.

@riccardo-forina I've moved regions check logic in machine code. Please have a look.

PR #197, opened 2022-03-09
feat: Connection tab in Kafka instance drawer

HemaHG:
Moving Connection tab content in the Kafka instance drawer
Related to Jira task https://issues.redhat.com/browse/MGDSTRM-7807

Screenshot from 2022-03-09 12-34-55

Where to find stories
Kafka> kafka instance drawer > kafka connection tab > connection tab

- [ ] test is pending


locales/en/common.json
```diff
@@ -9,6 +9,7 @@
   "delete": "Delete",
   "developmentPreview": "Development preview ",
   "developmentPreviewTooltip": "Development preview releases provide early access to a limited set of features that might not be fully tested. Users should not use development preview software in production or for business-critical workloads.",
+  "drawer_resource_tab_body_description_1": "Your application or tool initially connects to the Kafka instance using the bootstrap server and authenticates with your service accounts credentials.",
```
riccardo-forina, line 12: This doesn't feel like a common string

locales/en/kafka.json
```diff
@@ -17,12 +18,14 @@
     "consumer_id": "Consumer ID",
     "consumergroup_successfully_deleted": "Successfully deleted consumer group {{name}}",
     "current_offset": "Current offset",
+    "dead_state": "Dead",
```
riccardo-forina, line 21: For these states maybe group them in a states key? Perhaps at the root level of the kafka keys, I feel like state is something we might end up showing in other places not just into the drawer

locales/en/kafka.json
```diff
@@ -75,5 +75,17 @@
   "topic": {
     "topic": "Topic"
   },
-  "will_expire": "This instance will expire in <time>{{time}}</time>"
+  "will_expire": "This instance will expire in <time>{{time}}</time>",
+  "serviceAccount": {
```
riccardo-forina, line 79: I'd move this to a service-account.json file. SAs are used across MAS, which isn't limited to Kafka

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
```diff
@@ -0,0 +1,132 @@
...
+  externalServer,
+  tokenEndPointUrl,
+  linkToServiceAccount,
+  linkToAccessTab,
+  showCreateServiceAccountModal,
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className="mas--details__drawer--tab-content">
```
riccardo-forina, line 37: Do we have a corresponding css class somewhere?

riccardo-forina (approved):
It looks good, I left a few comments

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
HemaHG: you are right, css was not added. added it now, the same css is used in KafkaInstanceDrawer.tsx file so imported the it in that file also

locales/en/kafka.json
HemaHG: these strings are not related to this PR, maybe due to formatting strings being shuffled, can I group them in another PR?
Also, there are some strings related to the service account which is defined in kafka.json, can I move those to another PR

riccardo-forina: Oh ok sure

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
riccardo-forina: I see. We really should not import the same css more than once tho, because it will duplicate the same css code for every time we imported it. I'd suggest moving that css code to the drawer component's css file (assuming we have one, else let's create it) and keep the classes as they are in the other components.

locales/en/service-account.json
```diff
@@ -0,0 +1,12 @@
...
+  "service_accounts": "Service Accounts",
+  "service_accounts_small": "Service accounts",
+  "create_service_account_to_generate_credentials": "Create a service account to generate credentials. Manage the service accounts you create on the",
+  "create_service_account": "Create service account",
+  "current_instance": "Accounts do not have access to this instance unless granted under the instance’s",
+  "access_tab": "Access tab.",
+  "alter_allow" : "The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.",
+  "sasl_plain_description": " Use your service account credentials (Client ID and Client secret) as the user name and password to authenticate your client with the Kafka instance.",
+  "sasl_oauthbearer_description": "Use your service account credentials and a token endpoint URL to authenticate your client with the Kafka instance.",
+  "sasl_oauthbearer_popover_content": "OpenShift Streams for Apache Kafka uses the industry-standard OAuth protocol via the SASL/OAUTHBEARER provided by Apache Kafka."
```
jgiardino, line 11: The three strings above are not related to service accounts:
`sasl_plain_description`
`sasl_oauthbearer_description`
`sasl_oauthbearer_popover_content`

Should these be moved to `"connection_tab": { ... }` in kafka.json instead?

locales/en/common.json
```diff
@@ -25,5 +26,14 @@
...
-  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for."
+  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for.",
+  "bootstrap_server": "Bootstrap server",
+  "page": "Page",
+  "sasl_plain": "SASL/PLAIN",
+  "token_endpoint_url": "Token endpoint URL",
+  "more_info_about_sasl_oauthbearer": "More information about SASL/OAUTHBEARER",
+  "sasl_oauthbearer": "SASL/OAUTHBEARER",
+  "recommended": "Recommended",
+  "authentication_method": "Authentication method"
```
jgiardino, line 37: The strings you're adding here are contents that display on the Connection tab. Do we have guidelines that describe what makes a string a `common` string? Something seems odd about separating strings like
`bootstrap_server` and `bootstrap_server_description`, or `sasl_plain` and `sasl_plain_description`. 

Should these be moved to "connection_tab": { ... } in kafka.json instead?

locales/en/common.json
```diff
@@ -25,5 +26,14 @@
   "type_value_to_confirm_plain": "Type {{value}} to confirm",
   "unexpected_error": "There was a problem processing the request. Please try again.",
   "updated": "Updated",
-  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for."
+  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for.",
+  "bootstrap_server": "Bootstrap server",
+  "page": "Page",
```
jgiardino, line 31: Where this text displays in the UI, it shouldn't have a capital "P"
image

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <TextContent className="pf-u-pb-sm">
+        <Text component={TextVariants.h3} className="pf-u-mt-xl">
+          {t("service-account:service_accounts_small")}
+        </Text>
+        <Text component={TextVariants.small}>
+          {t("service-account:create_service_account_to_generate_credentials")}{" "}
+          <Link to={linkToServiceAccount} data-testid="tableStreams-linkKafka">
+            {t("service-account:service_accounts")}
+          </Link>{" "}
+          {t("common:page")}.
```
jgiardino, line 69: This structure is problematic for translation. When we do eventually send strings for translation, we want to send complete strings, like 
> Create a service account to generate credentials. Manage the service accounts you create on the [Service Accounts](https://619e1921b3e38b003afeb2be-lwmiwzufiw.chromatic.com/to-service-account) Page.

But with the way the strings are broken up here, a translator would receive these as separate strings:

- > Create a service account to generate credentials. Manage the service accounts you create on the
- > Service Accounts
- > page.

This is likely to result in poor translations. 

I'm not super familiar with the translation package that we're using. Is it possible to use a structure like this one: https://github.com/redhat-developer/app-services-ui-components/blob/main/locales/en/common.json#L25
e.g. `"Create a service account to generate credentials. Manage the service accounts you create on the {{pageName}} page."` with `"Service Accounts"` defined as a separate string that's passed in for `{{pageName}}`?

jgiardino:
@HemaHG Your updates look great! I left a few minor comments regarding translation strings, but I'm not sure that I was very helpful on that.

HemaHG:
@jgiardino, I followed the kas-ui.json file in the kas-ui repository and grouped the strings according to that, but I agree with your comment and will create a separate group for connection tab strings and place them in kafka-ui.json file

HemaHG:
@jgiardino, regarding the review comment https://github.com/redhat-developer/app-services-ui-components/pull/197#discussion_r823168823. I don't think it's not possible to structure all these strings into one string

HemaHG:
@riccardo-forina regarding the review comment https://github.com/redhat-developer/app-services-ui-components/pull/197#discussion_r822984750,there is no drawer component's css file and I think the css file I created is applicabel for kafkainstancedrawer component

HemaHG:
Thanks for the review @jgiardino, I have addressed all the review comments, could you please have a look

locales/en/kafka.json
```diff
@@ -4,10 +4,29 @@
...
   "connection_rate": "Connection rate",
   "connection_rate_value": "up to 100 connections/second",
+  "connection_tab": {
+    "access_tab": "Access tab.",
+    "alter_allow": "The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.",
+    "authentication_method": "Authentication method",
+    "bootstrap_server": "Bootstrap server",
+    "bootstrap_server_description": "Establish the initial connection between your client and Kafka instance.",
+    "create_service_account_to_generate_credentials": "Create a service account to generate credentials. Manage the service accounts you create on the <value>Service Accounts</value> page",
+    "current_instance": "Accounts do not have access to this instance unless granted under the instance’s",
```
jgiardino, line 14: This is another case where we should provide the complete statement for translation, but it's currently broken up into separate strings.

Can you follow the example you used for `create_service_account_to_generate_credentials` above to combine the following strings:
* `access_tab`
* `current_instance`

so that we have something like this?:
> `Accounts do not have access to this instance unless granted under the instance’s <value>Access tab</value>.`

**Note** that I moved the period (.) to be after the `</value>` so that it's not included as part of the link.

locales/en/kafka.json
```diff
@@ -4,10 +4,29 @@
...
+    "drawer_resource_tab_body_description_1": "Your application or tool initially connects to the Kafka instance using the bootstrap server and authenticates with your service accounts credentials.",
```
jgiardino, line 15: The last part of this statement should be:
> with your service account credentials.

i.e. remove the "s" at the end of "account"

locales/en/kafka.json
```diff
@@ -4,10 +4,29 @@
   "clientID": "Client ID",
   "connection_rate": "Connection rate",
   "connection_rate_value": "up to 100 connections/second",
+  "connection_tab": {
+    "access_tab": "Access tab.",
+    "alter_allow": "The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.",
+    "authentication_method": "Authentication method",
+    "bootstrap_server": "Bootstrap server",
+    "bootstrap_server_description": "Establish the initial connection between your client and Kafka instance.",
+    "create_service_account_to_generate_credentials": "Create a service account to generate credentials. Manage the service accounts you create on the <value>Service Accounts</value> page",
```
jgiardino, line 13: There should be a period (.) at the end of the last statement:
> ...`the <value>Service Accounts</value> page."`

jgiardino:
Thanks for making updates! I made a few comments on some really minor text string updates needed. Otherwise this looks good to me. 🎉

HemaHG:
@jenn I noticed that they are 2 dots(..) at the end of the 

> Accounts do not have access to this instance unless granted under the instance’s [Access tab](http://localhost:6007/to-service-account). The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.. 
Do we really need 2 dots? could you please check 

image

jgiardino:
> @Jenn I noticed that they are 2 dots(..) at the end of the

>> Accounts do not have access to this instance unless granted under the instance’s Access tab. The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access..

> Do we really need 2 dots? could you please check

@HemaHG - I didn't notice the two dots, but you're correct, there should only be one dot at the end of a statement.

> I have changed the code to use a single dot and there were two separate strings for

>> Accounts do not have access to this instance unless granted under the instance’s Access tab. and

>> The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access..

> combined both the strings, please let me know your thought on this.

I'm not sure if we should combine all statements in a paragraph in one translation string, or have separate translation strings per statement. @riccardo-forina, do you have any advice for this?

jgiardino:
Right now, we display these two statements in one paragraph in the UI:

> Accounts do not have access to this instance unless granted under the instance’s Access tab. 

> The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.

The first statement is duplicated for a similar chunk of UI for Service Registry, but the second statement is different, because the permissions needed are different. So for cases like that, do we use the same string for both Kafka and SR where they are identical? 

I'm not familiar enough with how translators work, aside from they get a list of strings, and provide translations for them. Depending on the tool, these strings could be adjacent in their list, but there's a possibility they aren't. If we provide duplicate strings in separate paragraphs, I think there's a risk that the person translating would not see that these are duplicates and therefore there's a potential for inconsistent translations.

jgiardino (approved):
LGTM!

PR #199, opened 2022-03-10
fix: update kafka details info on drawer and add stories

ajaypratap003:
These changes will be display in storybook only, not in production code. These changes will  move in production when instance info api available.

https://issues.redhat.com/browse/MGDSTRM-7506

image

image


src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.stories.tsx
```diff
@@ -17,20 +17,45 @@ const Template: ComponentStory<typeof KafkaDetailsTab> = (args) => (
...
-export const NormalInstance = Template.bind({});
-NormalInstance.args = {};
+const instanceInfo = {
+  ingress: "up to 50 MiB/second",
+  egress: "up to 50 MiB/second",
+  storage: "up to 1000 GiB",
+  maxPartitions: "up to 1500",
+  connections: "up to 2000",
+  connectionRate: "up to 100 connections/seconds",
+  messageSize: "up to 1 MiB",
```
riccardo-forina, line 27: The passed value should be just the numeric part. The string is responsibility of the UI and will have to be handled by the i18n framework

src/utils/kafka.ts
```diff
@@ -0,0 +1 @@
+export type InstanceType = "eval" | "standard";
```
riccardo-forina, line 1: Rename to `types.ts` please

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.stories.tsx
ajaypratap003: This data will be come through api. Not sure in which format api will return the data. If we are sure that api will return the data in numeric format  then  we can do changes now or maybe later we can change it when api will be ready. Anyways these changes are available storybook components  only, not in production code.

riccardo-forina: It will be numbers

src/utils/types.ts
```diff
@@ -0,0 +1 @@
+export type InstanceType = "eval" | "standard";
```
riccardo-forina, line 1: Sorry I didn't notice this earlier, this file should be in the folder of the feature. So `src/Kafka/KafkaInstanceDrawer/types.ts`

Given how it's used in a single place tho, is just make this type local to the component using it, no need to make it live in a single file or export it.

riccardo-forina: We will have one for the service registry then. The two things are semantically different things even if they have the same values.

jgiardino:
These updates look great! There are only a couple of things I noticed that didn't match the [design document](https://docs.google.com/document/d/1OARqcc1-RPq9B7XOqCSNivOQKmiXG_dbVMcI6ZCghrw/edit#heading=h.k4rzh13jc9zz).

1. For the standard example, there should also be a *Size* property that displays
2. All of the properties that we had before this update should have a new order as shown in the design document.

ajaypratap003:
> These updates look great! There are only a couple of things I noticed that didn't match the [design document](https://docs.google.com/document/d/1OARqcc1-RPq9B7XOqCSNivOQKmiXG_dbVMcI6ZCghrw/edit#heading=h.k4rzh13jc9zz).
> 
> 1. For the standard example, there should also be a _Size_ property that displays

We have size. you need to check latest story build.
 https://619e1921b3e38b003afeb2be-ylvdjcjxqy.chromatic.com/?path=/story/kafka-kafka-instance-drawer-kafka-details-tab--standard-instance-created
> 2. All of the properties that we had before this update should have a new order as shown in the design document.

I can't see new border in design doc. can you please paste design screenshot.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -10,13 +10,27 @@ import {
...
+
 type KafkaDetailsTabProps = {
   id: string;
   createdAt: Date;
   updatedAt: Date;
   expiryDate?: Date;
   owner: string;
   region: string;
+  instanceType: InstanceType;
+  size?: number;
```
riccardo-forina, line 23: Should this be optional? None of the other options are.

jgiardino: `size` should not display for trial instances, and `expiryDate` should not display for standard instances. Those should be the only properties that are different between trial and standard instances.

riccardo-forina: Ah, I didn't know that and it wasn't obvious looking at just the code. I'd make two separate components then, one for trial instances and one for standard instances. This way it's obvious what it is the required data for both cases.

ajaypratap003:
> > > All of the properties that we had before this update should have a new order as shown in the design document.
> 
> > I can't see new border in design doc. can you please paste design screenshot.
> 
> I'm not referring to a visual "border", but rather the sequence of properties. The sequence you included does not match the sequence shown in the design document.

ok. I didn't notice order. I've fixed. Please have a final look so that we can merge this PR.

riccardo-forina:
I also left a comment that needs to be addressed

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -42,15 +66,80 @@ export const KafkaDetailsTab: FunctionComponent<KafkaDetailsTabProps> = ({
...
+            <>
+              {renderTextListItem(
+                t("common:type"),
+                instanceType === "eval"
+                  ? t("common:trial")
+                  : t("common:standard")
+              )}
+              {instanceType === "standard" &&
+                renderTextListItem(
+                  t("size"),
```
jgiardino, line 79: I don't see where this string is defined, but it currently displays as "size". 
Instead it should be "Size" with a capital S.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -42,15 +66,80 @@ export const KafkaDetailsTab: FunctionComponent<KafkaDetailsTabProps> = ({
...
+          {renderTextListItem(t("common:id"), id)}
+          {renderTextListItem(t("common:owner"), owner)}
+          {renderTextListItem(
+            t("common:time_created"),
+            format(createdAt, "PPPP,p")
+          )}
+          {renderTextListItem(
+            t("common:time_updated"),
+            format(updatedAt, "PPPP,p")
+          )}
```
jgiardino, line 137: Currently in console.redhat.com, these values display as:
image

But the example in storybook for this PR shows this:
image

The property labels are what I provided in the design, but another difference between these two is the way the date and time values are combined.

Currently we use just a space `...2022 6:00 PM`, but this update replaces the space with a comma not followed by a space`...2022,6:00PM`

Now that I see the comma, it does make sense to include it, but can you add a space after the comma so that it displays like this instead?:
> Tuesday, March 8, 2022, 6:43 AM

jgiardino:
Thanks for fixing the order, @ajaypratap003! It matches what I was expecting. I did notice a couple of really minor things that would be good to address. The rest looks good to me.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
ajaypratap003: done

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
ajaypratap003: done

jgiardino (approved):
Thanks for making the updates!

jgiardino:
Today I noticed a statement in this future epic brief about not displaying terms like eval, standard, production to users. I raised a question there and am waiting for feedback. Depending on the replies, we might pull the "Type" property out, and we might add "trial" to the inline alert (i.e. "This trial instance will expire in {x} hours {y} minutes.")

Just wanted to mention that in case we want to leave this open to make any changes before merging.

locales/en/common.json
```diff
@@ -25,5 +25,11 @@
   "type_value_to_confirm_plain": "Type {{value}} to confirm",
   "unexpected_error": "There was a problem processing the request. Please try again.",
   "updated": "Updated",
-  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for."
+  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for.",
+  "type": "Type",
+  "trial": "Trial",
+  "standard": "Standard",
```
jgiardino, line 31: Based on recent discussions, we will not include Type in the UI.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -10,13 +10,27 @@ import {
...
+import { InstanceType } from "../utils";
+
 type KafkaDetailsTabProps = {
   id: string;
   createdAt: Date;
   updatedAt: Date;
   expiryDate?: Date;
   owner: string;
   region: string;
+  instanceType: InstanceType;
```
jgiardino, line 22: Based on recent discussions, we will not include Type in the UI.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -42,15 +66,80 @@ export const KafkaDetailsTab: FunctionComponent<KafkaDetailsTabProps> = ({
       <DetailsTabAlert expiryDate={expiryDate} />
```
jgiardino, line 66: Based on recent discussions, instead of including the following information in the description list:
> **Type**    Trial

this alert should be updated to say:
> This trial instance will expire in {x} {unit} {y} {unit}

jgiardino:
There was further feedback in [this discussion thread](https://docs.google.com/document/d/1RxTRpjgA7t6OSBRBI7MjmMA21cvFCBbs6VZuRqPI6o4/edit?disco=AAAAWT9Rrao&usp=suggestion_email_discussion&ts=6239dfe2&usp_dm=false) on the epic brief I mentioned yesterday about not wanting to use terminology like "standard" in the UI. 

Based on the discussion, I updated the [design document](https://docs.google.com/document/d/1OARqcc1-RPq9B7XOqCSNivOQKmiXG_dbVMcI6ZCghrw/edit?pli=1#) with these changes:
* Removed **Type** from the description list.
* Added "trial" to the inline alert that communicates expiration.

I've made some comments inline related to these updates.

jgiardino (approved):
LGTM! Thanks for making these updates.

ajaypratap003:
Thanks for reviewing PR and suggestions.
Merging this PR.  Feel free to add comments if missing anything, will be covered in next PR.

PR #200, opened 2022-03-10
fix: create modal stories

ajaypratap003:


riccardo-forina:
Thanks Ajay, I found we had a few edge cases we weren't covering, I extended the stories and clarified what the outcome should be for them.

riccardo-forina:
@jgiardino I renamed the stories to follow the `On Form Load` convention we used for the other stories. Please check the following:

* Some Regions Disabled On Form Load
* All Regions Disabled On Form Load
* No Regions Returned From Api For A Provider On Form Load
* No Regions Returned From Api On Form Load

I also deprecated the `trial-unavailable` state, it's been replaced by `instance-unavailable`. There is also a new state, `regions-unavailable` that we can use when all regions are disabled, or we get no regions from the API call (see above stories).

We are showing the same message for both the `instance-unavailable` and `regions-unavailable` cases. See [here](https://github.com/redhat-developer/app-services-ui-components/pull/200/files#diff-091dedd38534f829fdcedd332e5be1459e96f28ad2d9ea95b077e7182efe912aR88-R89)
I think the message displayed works for both cases, but if you want to differentiate them we can do that.

riccardo-forina:
Oh, one last thing. All stories now use `quota` as the default instance availability unless the story is specifically about trial instances.

jgiardino (approved):
Thanks for making these updates!
And thanks for the explanation, @riccardo-forina. That helped with reviewing this PR.

For the current implementation, everything looks correct.

For the next set of updates we are planning, I would like work with our content designer to refine some of the messaging we have.

riccardo-forina (approved):

PR #209, opened 2022-03-15
chore: Add resource type select component to storybook

suyash-naithani:


src/Kafka/ManageKafkaPermissions/components/ResourceType.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+      story: `A user can select a valid resource type, the initial select value is undefined with placeholder text as Select type `,
+    },
+  },
+};
+export const InitialState = Template.bind({});
+InitialState.args = { initialOpen: true };
+InitialState.parameters = {
+  docs: {
+    description: {
+      story: `A user can select a valid resource type and see their action being loged in the actions tab `,
```
riccardo-forina, line 58:
```suggestion
      story: `A user can select a valid resource type and see their action being logged in the actions tab `,
```

src/Kafka/ManageKafkaPermissions/components/ResourceType.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onChangeValue={setResourceTypeValue}
+        initialOpen={true}
+      />
+    </Modal>
+  );
+};
+InteractiveExample.parameters = {
+  docs: {
+    description: {
+      story: `A user can select a valid resource type, the initial select value is undefined with placeholder text as Select type `,
```
riccardo-forina, line 49:
```suggestion
      story: `A user can select a valid resource type. A placeholder is shown until an option is selected`,
```

riccardo-forina (approved):

jgiardino:
1. Error validation should not display in the form elements until until after form submit for missing required values. The only reasons we would display error validation before form submit is if the user enters a value that we know at that time is not valid.
    
    image

2. I don't see a difference between the two stories **Initial State** and **Interactive Example**. Can you clarify how these are different?
3. There are some form components that are rendered inside of a modal and one—Permissions Dropdown—that is not. 
    * Should all form components provide both examples? I.e. an example in the context of the modal and an example of just the component? This would let us see the documentation that should display in the Docs view (this isn't visible if a modal is open)
    * When I look at the Permissions Dropdown component that's not in a modal, the menu behaves as expected in that it automatically changes direction depending on available space below the toggle. I don't see this behavior for the select menus that are in the modal. 
4. Regarding the following question raised in chat:
    > Do we need a validation text under every select component in assign permissions? In this PR, we are not showing the text but using showing a validation error on the select itself. Let me know what you think about this

    * Every form element has a potential to show validation. But validation should be included part of a `<FormGroup>` component. I know we ran into issues in the past with styles not rendering as expected when we included form inputs inside `<FormGroup>` without also wrapping them inside a `<Form>`. I'm not sure what the best way to handle cases like this should be. Do we want components that include the form input inside a form group along with the validation messages? My only preference is we do what makes these components easier to assemble and consume in the product, even if that means that the styles do not display as expected in some of our stories. We still have the full example that shows everything together.

suyash-naithani:
@jgiardino regarding your questions,
1. That sounds good, we can remove the need to have validation in this component, with the parent component checking if the value is correct or not on submit press. @riccardo-forina  would that require us passing an additional value to this component for validation? 
2. The interactive example, a value can be selected and shown in the select component, but initial state will not update the view, but the updated value will be logged in the actions tab. This is best seen in canvas tab imo, the docs tab can sometimes fail to update when navigating to different stories  
3. I'll let @riccardo-forina  take this one 
4. We can wrap this in a formGroup. We'll have to figure out the best way for this, either it be in this component or it's parent. Since we will only check for validation of this component on submit click, we can probably have the form group outside of this component

src/Kafka/ManageKafkaPermissions/components/ResourceType.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+      disableFocusTrap={true}
+    >
+      <Form>
+        <ResourceType {...args} />
+      </Form>
+    </Modal>
+  );
+};
+
+export const InitialState = Template.bind({});
```
riccardo-forina, line 29:
```suggestion
export const WorksWithinModals = Template.bind({});
```

This story is more about being sure the dropdown opens ok when in a modal than anything else.
I'd make it more clear with a different name and an updated description.

A story about being able to set an initial value could be named `WithInitialValue`, configured with an initial value other than `undefined`

riccardo-forina (approved):

PR #211, opened 2022-03-16
feat: Create table for consumer group

HemaHG:
Part of Jira task https://issues.redhat.com/browse/MGDSTRM-7704
 
 move the code related to ConsumerGroup Table to storybook
 
image

where to find stories
kafka > consumer groups > components > ConsumerGroupTable


HemaHG:
HI @riccardo-forina, I have made the components independent of Kafka SDKs.could you please review the PR

src/Kafka/ConsumerGroups/components/ConsumerGroupStateLabel.tsx
```diff
@@ -0,0 +1,23 @@
+import { useTranslation } from "react-i18next";
+import { ConsumerGroupState } from "../types";
+
+export const ConsumerGroupStateLabel = (state: ConsumerGroupState) => {
```
riccardo-forina, line 4: This uses hooks, it must be either a hook itself or a component. It's written like a component but you are not declaring it as a component. You are not calling this as a component either.

HemaHG:
@riccardo-forina regarding the review comment https://github.com/redhat-developer/app-services-ui-components/pull/211#discussion_r831133583, I have moved that code to the utils.ts file.let me know if it's good or should follow some other approach

riccardo-forina:
> @riccardo-forina regarding the review comment [#211 (comment)](https://github.com/redhat-developer/app-services-ui-components/pull/211#discussion_r831133583), I have moved that code to the utils.ts file.let me know if it's good or should follow some other approach

Moving the function doesn't solve the problem. You'll have to either transform that function to be an hook, or make it a component. I'd make it a component.

HemaHG:
Added test for all the scenario

HemaHG:
I have incorporated all the review comments and added tests, build is passing now could you please review the PR @riccardo-forina

src/Kafka/ConsumerGroups/components/ConsumerGroupTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+      return bValue.localeCompare(aValue);
+    });
+  }
+
+  const getSortParams = (columnIndex: number): ThProps["sort"] => ({
+    sortBy: sortBy,
+    onSort: (_event, index, direction) => {
+      setActiveSortIndex(index);
+      setActiveSortDirection(direction);
+      setSortBy({ index, direction });
```
riccardo-forina, line 65: Why 3 different states? Wouldn't `setSortBy` be enough?

src/Kafka/ConsumerGroups/components/ConsumerGroupTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  const tableRow = (consumer: ConsumerGroup) => {
+    if (consumerGroupByTopic) {
+      return [];
+    }
+    const actionResolver = (consumer: ConsumerGroup): IAction[] => [
+      {
+        title: t("common:delete"),
+        onClick: () => onClickDeleteModal(consumer),
```
riccardo-forina, line 76: It could be a good idea to have a jest test that checks for these callbacks, to see they are getting properly called with the right parameters. Do you know how to do that? Ping me in the chat in case you don't!

riccardo-forina (approved):
Fantastic, thanks! I left a comment to where to improve the test coverage even more, I'll let you decide if you want to do that.

riccardo-forina (approved):
Great job!

PR #213, opened 2022-03-17
chore: Add text input for resource prefix value

suyash-naithani:


src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.tsx
```diff
@@ -0,0 +1,38 @@
+import { FormGroup, TextInput, ValidatedOptions } from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+
+export type ResourcePrefixProps = {
+  value: string | undefined;
+  onChangeValue: (value: string | undefined) => void;
```
riccardo-forina, line 6:
```suggestion
  onChange: (value: string | undefined) => void;
```

Nitpicking. No need to repeat what's being changed, since it's the only change event triggered by this component (a component with more than one would be weird really)

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(await comp.findByDisplayValue("pet")).toBeInTheDocument;
+  });
+  it("should show a text input with validation error ", async () => {
+    const onChangeValue = jest.fn();
+    const comp = render(<InvalidInput onChangeValue={onChangeValue} />);
+    await waitForI18n(comp);
+    expect(
+      await comp.findByPlaceholderText("Enter prefix")
+    ).toBeInTheDocument();
+    expect(await comp.findByText("Required")).toBeInTheDocument();
```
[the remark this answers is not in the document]
jgiardino: That's a great question. 

If we want to show all the variations of validation for an input at the input component level, rather than a parent form level, does that include all the different validation messages that could display? E.g. "Required" and other messages about invalid characters? 

I think that could simplify navigation and make it easier to find specific examples (e.g. how does this component look if I type the wrong character?)

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.tsx
```diff
@@ -0,0 +1,38 @@
...
+      helperTextInvalid={t("common:required")}
+      fieldId={"resource-prefix-input"}
+    >
+      <TextInput
+        placeholder={t("resourcePrefix.placeholder_text")}
+        validated={invalid ? ValidatedOptions.error : ValidatedOptions.default}
+        value={value}
+        type="text"
+        onChange={onPrefixValueChange}
+        aria-label="Prefix value text input"
```
jgiardino, line 33: Whenever `aria-label` is defined it should always use a translated string. 

An aria-label also doesn't need to describe the thing—that's like putting the word "button" inside of a button 😉 Usually it just repeats the visible text that displays adjacent to the input.

Looking at the current implementation, this attribute doesn't provide user-friendly text. 

In this case the aria-label I would recommend is something like this: 
`"{selected resource type} prefix"` so that it would say "Topic prefix" or "Consumer group prefix", depending on the resource type they selected in the first column.

image

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.tsx
```diff
@@ -0,0 +1,38 @@
...
+    onChangeValue(value);
+  };
+
+  return (
+    <FormGroup
+      validated={invalid ? ValidatedOptions.error : ValidatedOptions.default}
+      helperTextInvalid={t("common:required")}
+      fieldId={"resource-prefix-input"}
+    >
+      <TextInput
```
jgiardino, line 27: The current implementation is using the [typeahead select component](https://www.patternfly.org/v4/components/select#typeahead). Is there a reason for using the `<TextInput>` here instead?

image

riccardo-forina: AFAIK, we are not actually suggesting any value there in the existing implementation. Considering that the designs also show a plain text input instead of a select, I asked @suyash-naithani to use a plain text input.

Did I miss that requirement?

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.tsx
```diff
@@ -0,0 +1,38 @@
...
+        placeholder={t("resourcePrefix.placeholder_text")}
```
jgiardino, line 28: We use a very similar component if the user chooses "Is" instead of "Starts with". Should this component support both options? 
image

Something to note is that the validation messages we display could vary depending on "is" vs "starts with". For example, ".." is a valid prefix, but not a valid name. 

So back to @riccardo-forina's previous question - maybe there should be the following components: 
* a typeahead select component without the form group
* a prefix typeahead inside a form group with all the validation messages
* a name typeahead inside a form group with all the validation messages

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.test.tsx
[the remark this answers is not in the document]
suyash-naithani: @riccardo-forina  we are showing the 'Required' validation error message in the resource type select as well.
@jgiardino  Currently there is no validation if a user enters a wrong character,only if he does not give any value and clicks submit will the validation error and error message 'Required' show up
I can add validations in the component itself if the user does enter a value that is not possible and show an error. If a user assigns permission to a topic resource which starts with 'cat' , and there is no topic in the instance that starts with cat, how would we treat that? Entering an invalid value could be treated the same way as entering a value that does not match any topics 
If we do want to show an error, in that case, what should the validation message be?

jgiardino:
Your updates look good!

There's an option with the typeahead select to allow the user to specify a new value and not select an existing value. The designs showing how we use this option are shown here: https://marvelapp.com/prototype/a5e2da1/screen/81950514

I think the prop you use to add this functionality is `createText=[translated string for "Use"]` 

I also pasted the relevant portion of that design below:
image

locales/en/manage-kafka-permissions.json
```diff
@@ -50,6 +50,11 @@
     "resource_type":"Resource type",
     "resource_type_selected":"{{value}} Resource type"
   },
+  "resourcePrefix":{
+    "placeholder_text_starts_with":"Enter prefix",
+    "placeholder_text_is":"Enter name",
+    "prefix_aria_label":"Select resource prefix"
```
jgiardino, line 56: A different aria-label should be used if the user is specifying a name. 

I'm pasting below a previous comment I provided that's related to this:
> Usually an aria-label should just repeat the visible text that displays adjacent to the input.
> ...
> In this case the aria-label I would recommend is something like this:
"{selected resource type} prefix" so that it would say "Topic prefix" or "Consumer group prefix", depending on the resource type they selected in the first column.

Is it possible to pass the selected resource type as part of the string here?
For example: 
* If user selects "Is" as the rule, then use "{resource type} name", e.g. "Topic name", "Consumer group name"
* If user selects "Starts with" as the rule, then use "{resource type} prefix", e.g. "Topic prefix", "Consumer group prefix"

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.test.tsx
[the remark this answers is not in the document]
jgiardino: Apologies for missing this comment before. Feel free to ping me if I don't respond to comments in your PR.

There is an open jira that describes the current issues we have related to how we handle error messages: https://issues.redhat.com/browse/MGDSTRM-7502

In that jira, there's [a document](https://docs.google.com/document/d/1ITKLM-IMLi7PpZdGAGKQ8lseQHr-PB62-JgYs-FsxoI/edit) that lists all the error validation messages aside from "Required" that should display.

Most of the reasons for displaying errors are due to users entering characters that can't possibly match a topic name. E.g. they enter invalid characters, or they specify that the topic name is "." or ".." which we don't allow. 

> Currently there is no validation if a user enters a wrong character,only if he does not give any value and clicks submit will the validation error and error message 'Required' show up

If you look at what exists in production, currently the validation messages display on submit for the original Add Permission row. The behavior is different for the new shortcut options as noted in the jira. The following is a screencap of what we have in production.

image



> If a user assigns permission to a topic resource which starts with 'cat' , and there is no topic in the instance that starts with cat, how would we treat that?

We would not block that. Showing the menu supports use cases like the following:
* User wants to enter a prefix "sales-" but as they type, they realize all the topic prefixes are actually "sale-" when they see the list of existing topic names. In this case, they don't ever select a menu option, but specify a new value "sale-" and select the option 'Use "sale-"' that displays in the menu.
* User wants to specify a name of a topic that exists, so they start typing part of the name until they see the option they want.

The reason why we would not block the user from specifying "cat" if no topic exists with that name is for the following use case:
* The user wants to give an account permissions to create a topic with the name "cat". In this case, the topic "cat" doesn't exist yet, so the user would type "cat" and see the menu option 'Use "cat"' display in the menu.

jgiardino:
Thank you for adding the "Use" option to the component. Looking at this again, the filtering doesn't behave as I would expect it to. 

If I start typing, the value I enter should be matched wherever that value occurs in an option. For example, I see these options are available
image

If I type test, I no longer see the option "topic-test"
image

Should this behavior be represented in the storybook examples, or is that handled somewhere else?


One other question, I saw that this moved to the shared section, but these examples are all specific to the manage permissions modal. I'm just curious what the reasons are for that. I'm still learning how we're organizing components in our storybook.

suyash-naithani:
@jgiardino  The filtering done here is handled automatically by PF Select, we have not added any custom filtering to this, which is why it does not filter 'topic-test' when we type 'test'. This behaviour can be changed by adding custom filtering. 
We decided to add this to a shared component because of the nature in which typeahead suggestions are shown, even though right now it's only used here, we might need it in the future when we want to display suggestions in a typeahead that are fetched from an api call

package.json
```diff
@@ -78,6 +78,7 @@
     "@types/jest": "27.4.1",
     "@types/react": "17.0.43",
     "@types/react-dom": "17.0.14",
+    "@types/lodash.debounce": "4.0.6",
```
riccardo-forina, line 81: There is no `package-lock.json` in this PR, I think there should be since you added things to the package.json

src/utils/useDebounce.ts
```diff
@@ -0,0 +1,21 @@
+import debounce from "lodash.debounce";
```
riccardo-forina, line 1: We can remove this file now since we are not using it

riccardo-forina (approved):

PR #216, opened 2022-03-18
feat: add size field and update information details layout

ajaypratap003:
https://issues.redhat.com/browse/MGDSTRM-7507

Note: Size filed is available only for storybook  due to not integrate api. 
Some changes will move in production code Information section layout change etc. 

https://619e1921b3e38b003afeb2be-glctroqsbr.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--trial-available-on-form-load

https://619e1921b3e38b003afeb2be-glctroqsbr.chromatic.com/?path=/story/kafka-create-kafka-instance-create-kafka-instance--default


src/Kafka/CreateKafkaInstance/components/InstanceInfo.tsx
```diff
@@ -53,126 +59,163 @@ export const InstanceInfo: VoidFunctionComponent<
...
-                <TextListItem component={TextListItemVariants.dt}>
-                  {t("kafka:duration")}
-                </TextListItem>
-                <TextListItem component={TextListItemVariants.dd}>
-                  {t("create-kafka-instance:duration_value", {
-                    value: trialDurationInHours,
-                  })}
-                </TextListItem>
-              </GridItem>
+      <Card>
```
jgiardino, line 68: Can you add the `isFlat` prop to the cards. This removes the shadow and creates just a solid border.

src/Kafka/CreateKafkaInstance/components/InstanceInfo.tsx
```diff
@@ -53,126 +59,163 @@ export const InstanceInfo: VoidFunctionComponent<
...
+        <CardHeader>
+          <Text component={TextVariants.h3}>{t("instance_information")}</Text>
+        </CardHeader>
```
jgiardino, line 71: There are a few changes here that I think would be good to make:
* replace `<CardHeader>` with `<CardTitle component="h2">` and don't use the `<Text>` component. For example
  `<CardTitle component="h2">Details</CardTitle>`
  * With the modal title being an h1, these headings should really be h2. But continuing to use the `<TextContent>` would make the `<h2>` elements too big. 
  * This change would be applied here and in the next card.
* In the `<CardBody>`, replace the `<TextList>` component and children with a `<DescriptionList isCompact>` component with the children it uses. 
  * This buys us a little bit more space by being able to use the `isCompact` prop.

If all of these changes are made, then I don't think we need to wrap them in the `<TextContent>` component at all.

src/Kafka/CreateKafkaInstance/components/InstanceInfo.tsx
```diff
@@ -53,126 +59,163 @@ export const InstanceInfo: VoidFunctionComponent<
...
-          style={{ marginTop: "20px" }}
-          onClick={onClickQuickStart}
-        >
-          {t("create-kafka-instance:quick_start_guide_message")}
-        </Button>
-      </TextList>
+          </TextList>
+        </CardBody>
+      </Card>
+      <Card isCompact className="pf-u-mt-md">
```
jgiardino, line 202: Instead of using utility classes like this, we can place the `<Card>`s into a Stack layout with a gutter.

src/Kafka/CreateKafkaInstance/components/TrialKafkaForm.tsx
```diff
@@ -0,0 +1,198 @@
...
+          validated={regionValidation}
+        />
+        {isTesting && (
+          <>
+            <Text component={TextVariants.p} className="pf-c-form__helper-text">
+              {t("cloud_region_description")}
+            </Text>
+            <Button variant={ButtonVariant.link} isInline>
+              {t("learn_about_cloud_regions")}
+            </Button>
```
jgiardino, line 118: Ideally, the font size for this link would be the same size as the helper text. The only way to do that is by wrapping this button with the `className="pf-c-form__helper-text"` like you have above.

src/Kafka/CreateKafkaInstance/components/TrialKafkaForm.tsx
```diff
@@ -0,0 +1,198 @@
...
+            >
+              {t("streaming_unit")}
+            </span>
+          </div>
+          <Text component={TextVariants.p} className="pf-c-form__helper-text">
+            {t("trial_kafka_size_description")}
+          </Text>
+          <Button variant={ButtonVariant.link} isInline>
+            {t("learn_about_sizes")}
+          </Button>
```
jgiardino, line 193: Same comment as above...
Ideally, the font size for this link would be the same size as the helper text. The only way to do that is by wrapping this button with the `className="pf-c-form__helper-text"` like you have above.

jgiardino:
Thanks for making these updates so fast. I posted some suggestions inline for minor adjustments. I'd still like to spend some time reviewing all the examples and double-checking that things look as expected.

One thing that I noticed is in this example for **Loading Data**. The helper text and links should not display under Size and Cloud region. At this time, we don't know if the user gets a trial or standard instance, so we shouldn't display any text here until we know what type of instance they will get.

image

ajaypratap003:
> Thanks for making these updates so fast. I posted some suggestions inline for minor adjustments. I'd still like to spend some time reviewing all the examples and double-checking that things look as expected.
> 
> One thing that I noticed is in this example for **Loading Data**. The helper text and links should not display under Size and Cloud region. At this time, we don't know if the user gets a trial or standard instance, so we shouldn't display any text here until we know what type of instance they will get.
> 
> image

@jgiardino Thanks for the feedback. I've covered all the points, please ave a look and let me know if missing anything.

src/Kafka/CreateKafkaInstance/CreateKafkaInstance.css
```diff
@@ -9,6 +9,6 @@
 .mk--create-instance-modal__sidebar--content dt {
   font-size: var(--pf-global--FontSize--sm);
 }
-.mas-m-modalTop {
-  top: var(--pf-global--spacer--md);
+.mk--create-instance-modal__alert--message {
+  margin-top: var(--pf-global--spacer--md);
```
riccardo-forina, line 13: This is messing up the layout for stories when there is no message being shown. Take the "Default" story for example, the "Name *" label should be visually aligned with the right box.

image

riccardo-forina (commented):
I think this looks good, but I'm very concerned about introducing regressions since we don't have tests checking this component yet. 

Since it unfortunately already happened, I think we have to keep this on hold until we properly cover the implementation on main with tests, and then have the tests check these changes.

src/Kafka/CreateKafkaInstance/CreateKafkaInstance.css
jgiardino: I'm not sure I follow what the issue is, or understand how to reproduce it. 

I did notice that the flex layout somehow magically aligns the first line of text in the first flex item with the first line of text in the second flex item. For example, if you remove the padding from the card title so that it shifts up, the Name input also shifts up.

jgiardino:
In the **Error on form load** example, we should show the skeleton for the Details card, too:
image

jgiardino:
For **Generic error on form submit** I see the following from the preview in this PR:
image

But when I look at the current implementation, I see the following message. This matches what I expect to see. 
image

jgiardino:
In the **Name taken on form submit**, this is the message that displays. It is different from what displays in the current storybook and shouldn't change. 
image

jgiardino:
For the **Over quota on form load** example, I'm not seeing the modal reflect that use case. Was that something you were planning to handle for this PR? (i.e. the messages shown in this section of [the design document](https://docs.google.com/presentation/d/1U4j27K3L-9Ry0GPe-fcKoU48dCPlEJsfoXJUpldAY3U/edit?pli=1#slide=id.g118b16caf36_0_73))

src/Kafka/CreateKafkaInstance/CreateKafkaInstance.css
ajaypratap003: Why do you think this alignment issue is related to this css ? This css class used in bottom for inline alert message. 
image


I figure out this issue. Details header taking top padding due to `<CardTitle component="h2">` tag. 
`padding-top:var(--pf-c-card--first-child--PaddingTop);`

@jgiardino any suggestion for fixing alignment issue for Name filed with Details box. 
One fix is we can over wright `CardTitle` css class or we can use alternate of  CardTitle.

riccardo-forina: Right, this specific line is unrelated. It's the default alignment of the FlexItem that needs to be tweaked. I'll comment on the right place

jgiardino:
@ajaypratap003 I took another look at the storybook preview. The components for instance info look good. Overall, the layout of the Size input and helper text with links also look good. 

The details that seem off are the use cases and what messages display for the different use cases. But I know you mentioned that you're not including all of the updates for this PR, so it's hard for me to know what I should review now and what you plan to address later. 

Let me know how you want to proceed with this one.

Another update that I realized should be made is the title in the modal. The title should be "Create a Kafka instance" instead of "Create instance". As other services are onboarding and we're reviewing the titles across all of the services, this is the pattern that we are using for now. That's a change that you could make in this PR, too.

ajaypratap003:
> @ajaypratap003 I took another look at the storybook preview. The components for instance info look good. Overall, the layout of the Size input and helper text with links also look good.
> 
> The details that seem off are the use cases and what messages display for the different use cases. But I know you mentioned that you're not including all of the updates for this PR, so it's hard for me to know what I should review now and what you plan to address later.
> 
> Let me know how you want to proceed with this one.
> 
> Another update that I realized should be made is the title in the modal. The title should be "Create a Kafka instance" instead of "Create instance". As other services are onboarding and we're reviewing the titles across all of the services, this is the pattern that we are using for now. That's a change that you could make in this PR, too.

@jgiardino I've updated title as "Create a Kafka instance". 
In this PR, I'm considering visual changes related to this jira https://issues.redhat.com/browse/MGDSTRM-7507


1.     Add new field `Size` on create modal. 

2.     Update Instance information section on create modal

3.     Update alert message for trial users 

Rest message will be covered in next PR as dependent on API response. 
Please let me know if anything you want to cover in this PR. It would be easy if we can have merge this PR soon and open a new PR for API integration changes.

riccardo-forina:
Just to be sure we are on the same page, this will not be merged until #240 will, so to check this against the updated test suite.

ajaypratap003:
merging this PR and will open new PR due to jobs are not working.

PR #224, opened 2022-03-21
chore: ACL filters - Chip group options - Sim's Work

cfullam1:
Sim's work


src/ProofOfConcepts/Chips/ConsumerGroupLabel.js
```diff
@@ -0,0 +1,26 @@
+import "@patternfly/react-core/dist/styles/base.css";
```
riccardo-forina, line 1: You should never import PatternFly's css directly, those are already available for you in Storybook.

riccardo-forina (changes requested):
`patternfly.css` and `styles.css` should not be included. Also left a comment about importing PatternFly's CSS, it's something we don't need to (and should not) do.

simrandhaliw:
Thanks for opening up this PR on my behalf, Cliodhna! We can close this one now. I have opened a separate PR with the updated code.

src/ProofOfConcepts/Chips/ConsumerGroupLabel.js
simrandhaliw: Thanks for letting me know! I removed all unnecessary CSS imports in the new PR.

PR #232, opened 2022-03-23
chore: try to restore Chromatic build url in the action url

riccardo-forina:


riccardo-forina:
This is not needed, the Chromatic url in the actions is still there, we just have to scroll a bit

PR #236, opened 2022-03-23
chore: improve test coverage of CreateKafkaInstance and components

riccardo-forina:
Adds tests for all stories about `CreateKafkaInstance`

### DIFF
--- a/src/Kafka/CreateKafkaInstance/components/CloudProviderTiles.tsx
+++ b/src/Kafka/CreateKafkaInstance/components/CloudProviderTiles.tsx
@@ -1,37 +1,68 @@
-import { Skeleton, Tile } from "@patternfly/react-core";
+import {
+  FormSelect,
+  FormSelectOption,
+  SelectProps,
+  Skeleton,
+  Tile,
+} from "@patternfly/react-core";
 import { AzureIcon } from "@patternfly/react-icons";
 import AwsIcon from "@patternfly/react-icons/dist/esm/icons/aws-icon";
-import { Fragment, VoidFunctionComponent } from "react";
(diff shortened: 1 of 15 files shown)


riccardo-forina:
@jgiardino I noticed we had a few gaps around the aria roles of the alerts, I tried to update the code to better follow [the accessibility guidelines](https://www.patternfly.org/v4/components/alert/accessibility). I'm no expert about this, if you could give it a check it would be appreciated.

riccardo-forina:
This introduces a new helper for testing dialogs, `renderDialog`. It adds a wrapping div around stories with an id `root`. We can point to that id when writing the `appendTo` callback to make sure that the dialog properly renders in the test container.

src/Kafka/CreateKafkaInstance/components/CloudProviderTiles.tsx
```diff
@@ -1,37 +1,68 @@
...
+          key={id}
+          className={"mas--CreateKafkaInstance__CloudProviderTile"}
+          title={displayName}
+          icon={tiles[id]}
+          isSelected={value === id}
+          isDisabled={isDisabled}
+          onClick={() => onChange(id)}
+        />
       ))}
+      <FormSelect
```
riccardo-forina, line 45: Since we are using the `Tile` component as a form field, we need to make it behave like a form field. The trick I used here is to have a shadow (hidden) select component that mimicks what the show in the tiles, to make this accessible.

riccardo-forina:
Not sure why actions are not being run. I'm gonna close and reopen the PR. Sorry for the noise...

PR #249, opened 2022-03-29
chore: ACL filter chip group options

simrandhaliw:


cfullam1:
Lgtm @simrandhaliw :) Had some errors locally might sync at some stage to see if this is on my side only! I was wondering whether there was contents of the Kebab menu and the dropdown at the very left in the toolbar were to contain certain strings ?

cfullam1 (approved):

simrandhaliw:
> Lgtm @simrandhaliw :) Had some errors locally might sync at some stage to see if this is on my side only! I was wondering whether there was contents of the Kebab menu and the dropdown at the very left in the toolbar were to contain certain strings ?

Thanks for reviewing this PR, @cfullam1! The bulk-select dropdown items in this PR are just the ones that are provided by default from [this](https://www.patternfly.org/v4/components/toolbar#stacked-example) PF example. But during the productization of this code, we would want to apply the behavior as shown in this PF demo example: https://www.patternfly.org/v4/components/table/react-demos/#bulk-select and follow [these](https://www.patternfly.org/v4/guidelines/bulk-selection/#bulk-selection) PF guidelines. 

We have already implemented the kebab menu in the ACL page on the consoledot portal, and we won't be introducing any changes to it. The kebab menu in this PR (which shows some random dropdown items) was just to illustrate the placement of the menu in the toolbar.

PR #239, opened 2022-03-25
feat: Add connectors card and service category label

srambach:
Adds a new connectors card to the overview page per https://issues.redhat.com/browse/MGDCTRS-645
Also adds service labels to all cards on the overview page.

PR #256, opened 2022-03-31
Add connectors card and service category labels, take 2

srambach:
This is the re-do of https://github.com/redhat-developer/app-services-ui-components/pull/239 that was reverted to save for later. DO NOT MERGE until ready in May!

Adds a new connectors card to the overview page per https://issues.redhat.com/browse/MGDCTRS-645
Also adds service labels to all cards on the overview page.


locales/en/overview.json
```diff
@@ -23,6 +25,11 @@
   "rhosakMainText": "Streams for Apache Kafka enables IT development teams to capture, process, and stream real-time data across hybrid cloud environments.",
   "rhosakSecondaryText": "This service is fully hosted and managed by Red Hat for stream-based applications",
   "rhosakTitle": "Red Hat OpenShift Streams for Apache Kafka",
+  "rhocMainText": "Connectors provides pre-built integrations to enable connectivity from data, services, and systems to Red Hat OpenShift Streams for Apache Kafka.",
+  "rhocSecondaryText": "This service is hosted and managed by Red Hat as an add-on for managed",
+  "rhocSecondaryTextContinued": ", and is offered to Streams for Apache Kafka users.",
```
jgiardino, line 30: These contents look good. 

Should these two strings be combined into a single string? I'm not exactly sure which pattern is best for this scenario. We have a couple already that might work, but maybe @riccardo-forina can confirm which one he recommends:

1. https://github.com/redhat-developer/app-services-ui-components/blob/main/locales/en/kafka.json#L11
2. https://github.com/redhat-developer/app-services-ui-components/blob/main/locales/en/kafka.json#L38

srambach: Ah, I forgot there was a way to insert things into the i19n strings. I'll fix that.

riccardo-forina:
Closing this one since actions are not being run for some reason

PR #274, opened 2022-04-07
feat: Add kafka admin URL section in the connection details tab

HemaHG:
Add Kafka admin URL section in the connection details tab

https://issues.redhat.com/browse/MGDSTRM-7015

design doc - https://docs.google.com/document/d/1GeZnF8d8k9esakNSlWC_2FUVbZWLnNFHedracLMYFTw/edit

**Verification steps** 
kafka > kafka Instance Drawer > kafka Connect Tab > Connection Tab

### DIFF
--- a/src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
+++ b/src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
@@ -45,7 +45,7 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
         <Text component={TextVariants.small}>
           {t("kafka:connection_tab.drawer_resource_tab_body_description_1")}
         </Text>
-        <Text component={TextVariants.h3} className="pf-u-mt-lg">
+        <Text component={TextVariants.h3} className="pf-u-mt-xl">
           {t("kafka:connection_tab.bootstrap_server")}
         </Text>
         <Text component={TextVariants.small}>
@@ -57,19 +57,20 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
           <ClipboardCopy
             data-testid="drawerStreams-copyBootstrapURL"
             textAriaLabel={t("kafka:connection_tab.bootstrap_server")}
+            isReadOnly
           >
             {externalServer}
           </ClipboardCopy>
         )}
       </TextContent>
       <TextContent className="pf-u-pb-sm">
-        <Text component={TextVariants.h3} className="pf-u-mt-lg">
+        <Text component={TextVariants.h3} className="pf-u-mt-xl">
           {t("kafka:connection_tab.rest_api")}{" "}
           <Popover
             aria-label={t("kafka:connection_tab.rest_api_popover_label")}
             bodyContent={
-              <div>
-                <div>
+              <TextContent>
+                <p>
                   {
                     <Trans
                       i18nKey={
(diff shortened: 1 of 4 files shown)


src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
```diff
@@ -56,6 +62,69 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
...
+                      i18nKey={
+                        "kafka:connection_tab.rest_api_popover_content_2"
+                      }
+                      components={{
+                        value: <Link to={linkToLearnMore}></Link>,
+                      }}
+                    />
+                  }
+                </div>
+              </div>
```
jgiardino, line 96: The formatting of these contents would be nicer if we use the `TextContent` component with `p` elements instead of divs for each paragraph.

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
```diff
@@ -56,6 +62,69 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
...
+        </Text>
+        {isKafkaPending ? (
+          <Skeleton fontSize="2xl" />
+        ) : (
+          <ClipboardCopy
+            data-testid="drawerStreams-copyRestApiURL"
+            textAriaLabel={t("kafka:connection_tab.rest_api")}
+          >
+            {restAPIUrl}
+          </ClipboardCopy>
```
jgiardino, line 125: I'm just now noticing that we have one clipboard copy that is gray and one that is white. My guess is that all of these should be gray (i.e. read-only), but I can confirm that with stakeholders.
 
image
image

src/Kafka/KafkaInstanceDrawer/__snapshots__/kafkaConnectionTab.test.tsx.snap
```diff
@@ -86,20 +86,117 @@ exports[`ConnectionTab renders 1`] = `
         class="pf-c-content pf-u-pb-sm"
       >
         <h3
-          class="pf-u-mt-xl"
+          class="pf-u-mt-lg"
```
jgiardino, line 89: I would change this back to `-xl` to be consistent with the margin-top that's applied to all `h3` elements in this panel.

jgiardino:
Thank you for making these updates, @HemaHG! 

I just posted a comment in the epic with questions for the team. Hopefully the preview will help us get some answers to my questions. I also provided feedback on some minor things I noticed.

HemaHG:
@jenn Initially when the Kafka instance creation just started the bootstrap and token end URL will be blank, will the same apply to the Kafka URL section?

jgiardino:
> @Jenn Initially when the Kafka instance creation just started the bootstrap and token end URL will be blank, will the same apply to the Kafka URL section?

> please refer to the below story for how the bootstrap and token end URL will look when kafka instance creation is pending, I have applied the same condition for the Kafka URL section.

Good catch! Thanks for adding this. 

With PTO schedules, I probably won't get answers to everything this week, and even then, the contents that are TBD might be TBD until after summit. I was able to get some feedback from @wtrocki that the clipboard copy controls should be read-only. I think we can make that update in this PR, then revisit this with further content updates when we have more information to provide.

src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
jgiardino: Following @wtrocki's feedback, let's make all clipboard copy elements read-only (i.e. gray background).

HemaHG:
Only storybook is created for connection Tab, it's not yet integrated with the business logic so merging this PR will not affect the Production

riccardo-forina (approved):

PR #275, opened 2022-04-07
chore: Add validation stories for resource prefix

suyash-naithani:

> This PR adds stories for all the error validation messages when assigning resource prefix in acls a value

Addresses [issue#6502](https://issues.redhat.com/browse/MGDSTRM-7502)

**Verification steps** 

You can find the stories in src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.stories


src/shared/AsyncTypeaheadSelect/AsyncTypeaheadSelect.tsx
```diff
@@ -47,6 +51,12 @@ export const AsyncTypeaheadSelect: VFC<AsyncTypeaheadSelectProps> = ({
     undefined
   );
 
+  useEffect(() => {
+    if (submitted && required)
+      setValidation({ isValid: false, message: t("common:required") });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
riccardo-forina, line 57: Never shut down this error, it's a hint that something is fishy in your logic.
In this case, you don't need an effect at all, the validation can be derived by combining the temporary one of the typeahead - in the local state - and the one coming from the props.

src/Kafka/ManageKafkaPermissions/components/ResourcePrefix.stories.tsx
```diff
@@ -0,0 +1,120 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { AsyncTypeaheadSelect } from "../../../shared";
+import { validationCheck } from "../storiesHelpers";
+import { Form } from "@patternfly/react-core";
+import { userEvent, within } from "@storybook/testing-library";
+import { fakeApi } from "../../../shared/storiesHelpers";
+
+export default {
+  component: AsyncTypeaheadSelect,
```
riccardo-forina, line 9: I'd expect a `ResourcePrefix` component here.

src/Kafka/ManageKafkaPermissions/storiesHelpers.ts
```diff
@@ -0,0 +1,35 @@
...
+    return {
+      isValid: false,
+      message:
+        "Valid characters in a consumer group ID include letters (Aa–Zz), numbers, underscores ( _ ), and hyphens ( - ).",
+    };
+  if (filter.length > 32)
+    return { isValid: false, message: "Cannot exceed 32 characters" };
+
+  return undefined;
+};
```
riccardo-forina, line 35: This belongs to the component, it's not something used only for the stories of the component.

src/shared/AsyncTypeaheadSelect/AsyncTypeaheadSelect.test.tsx
```diff
@@ -13,33 +13,29 @@ const {
 
 describe("Async typeahead", () => {
   jest.useFakeTimers();
-  it("should render a select with async typeahead suggestions", async () => {
+  it("It should render an async typeahead", async () => {
```
riccardo-forina, line 16: Nitpick, no need to write "It" in the description since it's already there in the code

suyash-naithani:
@riccardo-forina  I pushed the changes that you suggested in your comments. Let me know what you think.

jgiardino:
Overall, your examples look really good! The issues I'm seeing in the storybook examples are really minor.

In [Invalid Topic Length](https://619e1921b3e38b003afeb2be-snnzephcgu.chromatic.com/?path=/story/kafka-manage-kafka-permissions-components-resource-prefix--invalid-topic-length), this case would only surface if the user selects the prefix rule "Is" and the resource type "topic" and types either "." or "..". The label of this group of components and this story don'f fully convey that. 
* If these stories are meant to support either rule—"Is"/`literal` or "Starts with"/`prefix`—then changing the group label "Resource prefix" to "Resource name" or "Resource literal/prefix" might help clarify this use case. 
* This story isn't related to max length, so changing the name from "Invalid topic length" to "Invalid topic name" would be good, too.

jgiardino:
One other issue I see is with the aria attributes defined on this select.

- The values for `aria-label` and `placeholder` assigned to the input to not match what I expect to see. Both should either include "name" or "prefix" because the user is either specifying a name or a prefix depending on the rule they selected. For example:
  - `aria-label="{{value}} prefix" placeholder="Enter prefix"`
  - `aria-label="{{value}} name" placeholder="Enter name"`
- The value specified for the `id` is not valid because ids should not include spaces. The id is important here because this is how we associate the error text with the form input, for users that are using a screen reader. 
  - Note that according to [PF react documentation](https://www.patternfly.org/v4/components/select#select), you _should_ be able to provide `aria-describedby="[id of helper text]"` on the Select component to make this association, but I'm not seeing this work as expected. I will open an issue for this to get fixed in PF, but this is also something that needs to be added on our end. I think we should add it now following their documentation while we're working on this item. 

image

locales/en/manage-kafka-permissions.json
```diff
@@ -16,22 +16,17 @@
     "all": "All",
     "alter": "Alter",
     "alter_configs": "Alter configs",
+    "aria_label": "Resource type operation value",
```
jgiardino, line 19: What is this string for?
`aria-label` is a common html attribute that can be appear for many different elements. Using this as an id for a string isn't specific enough. 

Including "value" isn't needed here. It's like including the word "button" as part of the button text.

jgiardino: Oh, ignore my first question. I expanded the context of this rule, and see it's nested under `"operations"`. 🤦‍♀️

locales/en/manage-kafka-permissions.json
```diff
@@ -16,22 +16,17 @@
...
-  "resource_prefix_rule": {
-    "is": "Is",
-    "starts_with": "Starts with",
-    "aria_label": "Resource prefix rule {{value}}",
-    "starts_with_description": "Refers to the prefix of the name in the API",
-    "is_description": "Refers to the literal name in the API"
+    "write": "Write"
   },
+  "pattern_type_literal": "",
+  "pattern_type_prefixed": "",
```
jgiardino, line 29: Why are these empty?

locales/en/manage-kafka-permissions.json
jgiardino: For my second comment, further down, I see examples where the aria-label is `"... {{value}}"`. Is that what's intended here?

locales/en/manage-kafka-permissions.json
```diff
@@ -48,7 +43,25 @@
...
+    "create_text": "Use",
+    "invalid_consumer_group": "Valid characters in a consumer group ID include letters (Aa–Zz), numbers, underscores ( _ ), and hyphens ( - ).",
+    "invalid_length": "Cannot exceed 32 characters",
+    "invalid_topic": "Valid characters in a topic name include letters (Aa-Zz), numbers, underscores ( _ ), periods ( . ), and hyphens ( - ).",
+    "invalid_topic_length": "A topic name must contain at least 3 periods (...) if periods are the only characters used",
+    "name_aria_label": "{{value}} name",
+    "placeholder_text_is": "Enter name",
+    "placeholder_text_starts_with": "Enter prefix",
+    "prefix_aria_label": "{{value}} prefix"
+  },
```
jgiardino, line 64: This looks correct. 

One minor suggestion—it might be easier to understand what's being defined here if we're consistent in what we call these. E.g. either use "name" and "prefix" or use "is" and "starts_with". They're referring to the same thing, so using a mix of the two could be confusing for someone else who's looking at this later.

For example:
`"aria_label_name_is"`
`"placeholder_name_is"`
`"aria_label_name_starts_with"`
`"placeholder_name_starts_with"`

Also, the examples are all using the `name` rule. Should we add one story that is using `prefix`?

locales/en/manage-kafka-permissions.json
```diff
@@ -48,7 +43,25 @@
...
+    "starts_with": "Starts with",
+    "starts_with_description": "Refers to the prefix of the name in the API"
+  },
   "resource_type_operation_filter_placeholder": "Filter by resource type operation",
+  "resourcePrefix": {
+    "create_text": "Use",
+    "invalid_consumer_group": "Valid characters in a consumer group ID include letters (Aa–Zz), numbers, underscores ( _ ), and hyphens ( - ).",
+    "invalid_length": "Cannot exceed 32 characters",
+    "invalid_topic": "Valid characters in a topic name include letters (Aa-Zz), numbers, underscores ( _ ), periods ( . ), and hyphens ( - ).",
+    "invalid_topic_length": "A topic name must contain at least 3 periods (...) if periods are the only characters used",
```
jgiardino, line 59: One other minor comment—
The string `invalid_length` is used when the user enters too many characters, but this message is not about length.
This message displays if the user specifies either "." or ".." when "Is" is selected. Maybe `invalid_topic_name` is more applicable here?

jgiardino:
Your updates look good! There are just a couple of minor things I noticed. I'm not sure if they need to be addressed as part of this PR or not. 

One question that I have is whether the storybook components are meant to be fully functional or not. When I look at the [component example on PF.org](https://www.patternfly.org/v4/components/select#typeahead), I'm able to click the down arrow 🔽  to open the menu. I don't see this behavior in the Resource Name components. I do see the menu, but only when I clear the value in the text input of the typeahead toggle.
@riccardo-forina - Would that behavior be expected at this level?

One minor thing I noticed when looking at [Invalid Topic Name](https://619e1921b3e38b003afeb2be-qvsbtkvkyq.chromatic.com/?path=/story/kafka-manage-kafka-permissions-components-resource-name--invalid-topic-name), the message displays immediately as the user types. 

This behavior is fine for other components like invalid characters or invalid length. If I type "$", there is nothing I can do to make that character valid. And if I type 1 too many characters, there's nothing I can do to make that character length valid. So displaying the message as soon as the user encounters these error conditions makes sense.

For the scenario where topic name cannot be "." or "..", this is a valid user flow:
1. User types ".."
2. Menu that lists topic names is filtered to only show the topic names that include ".."
3. User selects the topic "..bar" in the menu of filtered results

If we display an error message at step 1 above, we're making an assumption that could be false as illustrated with the above scenario. Instead, this is the flow where it's more appropriate to display the message.

1. User types ".."
2. Menu that lists topic names is filtered to show the topic names that include "..", with the last option being 'Use ".."'
3. User selects the option 'Use ".."'
4. User sees the error message below the input. 

I wouldn't consider this critical to address now, in that I don't see it as blocking the user, it's just a bit obnoxious. 
@riccardo-forina - What would you suggest for this one?

riccardo-forina:
> One question that I have is whether the storybook components are meant to be fully functional or not. When I look at the [component example on PF.org](https://www.patternfly.org/v4/components/select#typeahead), I'm able to click the down arrow 🔽 to open the menu. I don't see this behavior in the Resource Name components. I do see the menu, but only when I clear the value in the text input of the typeahead toggle. @riccardo-forina - Would that behavior be expected at this level?

I guess it depends. For low-level components like this one that depends on a parent component to deal with keeping the typed value, for example, it's a good idea to have a set of non-functional stories that represent exactly the state you want to test. It's also possible to provide a fully working example tho, and I think we do that sometimes. Given that this is a component that will be used inside a bigger feature and not really shared around, I'd make the fully-featured example only for the bigger feature and not for these smaller blocks, just to save time. This is the pattern I followed for the message browser for example.

> I wouldn't consider this critical to address now, in that I don't see it as blocking the user, it's just a bit obnoxious. @riccardo-forina - What would you suggest for this one?

I don't know. Can we have two levels of validations perhaps? One that's checked while typing that checks for syntactically incorrect values, and one that's run when the user blurs out of the component where we do a more context-specific validation?

jgiardino:
> I don't know. Can we have two levels of validations perhaps? One that's checked while typing that checks for syntactically incorrect values, and one that's run when the user blurs out of the component where we do a more context-specific validation?

I definitely want to utilize these levels of validation:

- checked while typing and displays immediately (e.g. invalid characters or length typed will always be invalid, so let's tell the user immediately)
- checked on form submit (e.g. we don't know yet if the user plans to submit a form without a required field until they actually submit the form, so let's tell them on form submit. Otherwise, we're being obnoxious and potentially obscuring help text) 

But in this example, there's a typeahead which adds another level between "typing" and "form submit" which this:
- checked after selecting an option — but really in this case, it's when the user selects the option 'Use "{value}"`. While the user is typing in the typeahead, they might be trying to filter or they might be typing the {value} they want to use. We don't know that until they make a selection in the menu.

riccardo-forina (approved):
Let's move this forward. We'll review any eventual gap within the scope of the PR that implements the whole dialog.

locales/en/manage-kafka-permissions.json
```diff
@@ -53,10 +53,14 @@
...
   "resourcePrefix": {
     "create_text": "Use",
-    "placeholder_text_is": "Enter name",
-    "placeholder_text_starts_with": "Enter prefix",
-    "prefix_aria_label": "Resource prefix",
-    "select_prefix_aria_label": "{{value}} prefix"
+    "invalid_consumer_group": "Valid characters in a consumer group ID include letters (Aa–Zz), numbers, underscores ( _ ), and hyphens ( - ).",
+    "invalid_length": "Cannot exceed 32 characters",
+    "invalid_topic": "Valid characters in a topic name include letters (Aa-Zz), numbers, underscores ( _ ), periods ( . ), and hyphens ( - ).",
+    "invalid_topic_name": "A topic name must contain at least 3 periods (...) if periods are the only characters used",
```
jgiardino, line 59: Please add a period "." at the end of this string.

jgiardino:
I took another look at the UI component and I just had one minor comment about text strings. For any of my previous comments, let's revisit them when we pull everything together in a full preview of the modal.

PR #286, opened 2022-04-11
feat: Add storybook for Consumer group reset offset component

HemaHG:

> Create a storybook for ConsumerGroup ResetOffset Modal component

**Verification steps** 
Screenshot from 2022-04-05 16-38-13

**where to find stories:**
kafka > Consumer Groups > components > Consumer Group Reset Offset
https://619e1921b3e38b003afeb2be-aldaurdobk.chromatic.com/?path=/story/kafka-consumergroups-components-consumergroupresetoffset--no-topic-selected

Tests are pending currently working on it


locales/en/common.json
```diff
@@ -31,5 +31,6 @@
   "type_value_to_confirm_plain": "Type {{value}} to confirm",
   "unexpected_error": "There was a problem processing the request. Please try again.",
   "updated": "Updated",
-  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for."
+  "we_cannot_find_the_page_you_are_looking_for": "We can’t find the page you’re looking for.",
+  "select": "Select"
```
riccardo-forina, line 35: Please run `npm run parse-i18n` when you add keys, so to have them sorted alphabetically and avoid conflicts in future PRs if someone runs that command.

src/Kafka/ConsumerGroups/ConsumerGroup.css
```diff
@@ -0,0 +1,7 @@
+#custom-offset-input {
+  width: 10rem;
+}
+
+.form-title {
```
riccardo-forina, line 5: This needs to be better namespaced.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,251 @@
...
+            >
+              <p>{t("consumerGroup.reset_offset_connected_alert_body")}</p>
+            </Alert>
+          )}
+        </StackItem>
+        <StackItem>
+          {isDisconnected && consumers.length > 0 && selectedTopic && (
+            <Stack hasGutter>
+              <StackItem>
+                <TableComposable
```
riccardo-forina, line 175: Can we try using the `ResponsiveTable` component we have in the shared components for this?

src/Kafka/ConsumerGroups/components/OffsetDropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+    ));
+  };
+
+  return (
+    <Dropdown
+      onSelect={onSelect}
+      dropdownItems={dropDownItems()}
+      isOpen={isOpen}
+      id="offset-dropdown"
+      aria-label="offset-select-dropdown"
```
riccardo-forina, line 52: This is not a proper label to present to a text reader. You should ask @jgiardino if in doubt about what to use here, but at the very least it should be the same label used for the wrapping `FormGroup`

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,251 @@
...
+        </Button>,
+        <Button variant="link" key={2} onClick={onClickClose}>
+          {t("common:cancel")}
+        </Button>,
+      ]}
+    >
+      <Stack hasGutter>
+        <StackItem>
+          <Form isHorizontal>
+            <FormGroup label="Consumer group" fieldId="horizontal-form-name">
```
riccardo-forina, line 125: All these hard-coded labels should be under translation.

src/Kafka/ConsumerGroups/components/TopicsDropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  const dropDownItems = topics.map((topic) => (
+    <DropdownItem key={topic}>{topic}</DropdownItem>
+  ));
+
+  return (
+    <Dropdown
+      id="topic-dropdown"
+      aria-label="topic-select-dropdown"
```
riccardo-forina, line 40: Same as above.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,251 @@
...
+    log_end_offset: t("consumerGroup.log_end_offset"),
+    offset_lag: t("consumerGroup.offset_lag"),
+    new_offset: t("consumerGroup.new_offset"),
+  };
+
+  return (
+    <Modal
+      variant={ModalVariant.large}
+      isOpen={isModalOpen}
+      aria-label={"Modal for resetting offset of Kafka consumergroup"}
```
riccardo-forina, line 104: Please, under translation.

riccardo-forina (changes requested):
I left a few comments, I think this is looking good but requires more cleanup.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
HemaHG: @riccardo-forina, I have a column for selecting the row in the table, but I don't think we can achieve that with the  `ResponsiveTable `

riccardo-forina: It should be supported, the `Th` you get from the `ResponsiveTable` is HOC of the `Th` you use here. All its features are supported.

Did you try and get some errors?

HemaHG: I need 
image selectable column, Not sure how to deal with this with the `ResponsiveTable`

riccardo-forina: I added an example here. I'll update it to be more readable eventually, but it should give you an idea of how to do this.

https://github.com/redhat-developer/app-services-ui-components/pull/377

HemaHG: As per your comment, https://github.com/redhat-developer/app-services-ui-components/pull/377#issuecomment-1139487866 says not to follow this approach, so for now can I use the normal table?

riccardo-forina: Correct

src/Kafka/ConsumerGroups/components/TopicsDropdown.tsx
jgiardino: My comments about using a Select component instead of Dropdown (and how to associate accessible text with the Select component) apply here as well.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
jgiardino: An `aria-label` isn't needed if the `title` prop is used on a modal. The PF Modal automatically uses the value provided for `title` as the accessible label for the modal.

jgiardino:
@riccardo-forina - @HemaHG was also asking about the Delete modal and if it should include a warning icon. For this delete modal, I'm following what PF recommends for [confirmation prompts](https://www.patternfly.org/v4/components/modal/design-guidelines#confirmation-dialogs) when the action is not destructive. Even though the user is deleting a consumer group, the action isn't considered destructive. The consumer group is created again as soon as it becomes active. 

An overview of design patterns I'd like to follow for our confirmation prompts (plus some updates needed to existing prompts), is available here: https://docs.google.com/document/d/1sNRSJtCpR9OHVvzA-D00WMlughX4XNP4RFThVuIdZ4A/edit?usp=sharing

src/Kafka/ConsumerGroups/components/OffsetSelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <Select
+      aira-label={t("consumerGroup.reset_offset_new_offset_label")}
+      toggleAriaLabel={t("consumerGroup.reset_offset_new_offset_label")}
+      id={"topic-select"}
+      onToggle={onToggle}
+      onSelect={onSelect}
+      variant={SelectVariant.single}
+      isOpen={isOpen}
+      placeholderText={t("common:select")}
+      width={200}
```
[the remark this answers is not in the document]
jgiardino: Can you set the width here to be 100% and then set `maxWidth="600px"` on the parent `<Form>`? 

Otherwise, if you set the width here, we would want something like width=100% and max-width=400px. I'm not sure if those are both available as props on the select.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,262 @@
...
+        </Button>,
+        <Button variant="link" key={2} onClick={onClickClose}>
+          {t("common:cancel")}
+        </Button>,
+      ]}
+    >
+      <Stack hasGutter>
+        <StackItem>
+          <Form isHorizontal>
+            <FormGroup label="Consumer group" fieldId="horizontal-form-name">
```
jgiardino, line 125: Using a `<Title>` component here doesn't seem appropriate.

One option is to just include `{groupID}` as the direct child of `<FormGroup hasNoPaddingTop>`. The `hasNoPaddingTop` removes the top padding so that the label lines up better with this text.

Another option is to use the read-only styles we have in the Kafka topic page. We have a custom class for `<FormGroup>` - `.kafka-ui-form-group--readonly` - which lets us use a read-only input to display read-only values in a similar visual style. 

image

P.S. I would definitely recommend the second pattern if we have a really long form where some properties are read-only and some are editable (like the edit page). But in this case, the user pre-selected the consumer group, so it's just here to provide additional context.

src/Kafka/ConsumerGroups/components/OffsetSelect.tsx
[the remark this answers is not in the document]
HemaHG: @riccardo-forina ```
All form inputs should either be the same length. Sometimes form inputs for values that are expected to be a few characters will be intentionally short, but I'm not sure that really applies to our case with the New offset input. You can see more details about form input width in these PF guidelines: https://www.patternfly.org/v4/components/form/design-guidelines#forms-on-a-page
I would rather have the form input slightly too long rather than too short and end up cutting off topic names that are long
I'm thinking that a max-width of 400px would be good, and a width of 100%. This way, if the modal is smaller than 400px, it displays at 100% instead of getting cut off by the modal.
``` These are the thoughts of @jgiardino on select input length and also she suggested taking your opinion on this, Could you please provide your feedback

File: src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx (hunk 9 above)
Comment (HemaHG) [reply src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx]: @jenn as per the @riccardo-forina comment, I removed the width in the select component, and also replaced the title component with Textinput, here is the updated story https://619e1921b3e38b003afeb2be-ypuxtqyvbq.chromatic.com/?path=/story/kafka-consumergroups-components-consumergroupresetoffset--reset-offset-when-offset-is-absolute

File: src/Kafka/ConsumerGroups/components/OffsetSelect.tsx (hunk 8 above)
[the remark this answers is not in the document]
Comment (jgiardino) [reply src/Kafka/ConsumerGroups/components/OffsetSelect.tsx]: That makes sense, @riccardo-forina! I also realize that the guidance from PF is specifically for a form on a page: 
> To accommodate lengthy input fields, the maximum suggested width is 880px. Use your best judgement based on the context and content of your form –– 880px is typically too wide for most use cases.

but now that I scroll down to the section about modals, it's recommending what you suggest above: 
> Forms in a modal should be horizontally centered with the modal background and fill the modal width, leaving 24px between the form content and the modal background edges.

File: src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx [hunk 10]
```diff
@@ -0,0 +1,274 @@
...
+              className="mas--ConsumerGroupResetOffset-form-group--readonly"
+              label={t("consumerGroup.reset_offset_consumer_group_label")}
+              fieldId="horizontal-form-name"
+            >
+              <TextInput
+                isReadOnly
+                type="text"
+                aria-label={t(
+                  "consumerGroup.reset_offset_consumer_group_label"
+                )}
```
jgiardino, line 134: Since a text input can use the native html `<label for="">`/`<input id="">` association for providing an accessible name, this is preferred over using `aria-label`. 

To use this method, you can make these changes:
- assign a unique `id` to this `<TextInput>` component )
- use the same value you give the `id` as the value for `fieldID` above
- remove the `aria-label` here

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,274 @@
...
+                      );
+                    })}
+                  </Tbody>
+                </TableComposable>
+              </StackItem>
+              <StackItem>
+                <Checkbox
+                  label={t("consumerGroup.reset_offset_accept")}
+                  aria-label="uncontrolled checkbox example"
+                  id="check-5"
```
jgiardino, line 263: This ID seems really generic, which raises the risk of a duplicate ID being used somewhere else on the page. Can you provide a more specific ID that is unique to this control?

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,274 @@
...
+                        </Tr>
+                      );
+                    })}
+                  </Tbody>
+                </TableComposable>
+              </StackItem>
+              <StackItem>
+                <Checkbox
+                  label={t("consumerGroup.reset_offset_accept")}
+                  aria-label="uncontrolled checkbox example"
```
jgiardino, line 262: `aria-label` should not be used here. The `<Checkbox>` automatically handles associating the label with the input using the `id`/`for` method.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+    topics: getTopics(),
+    consumers: consumer,
+  },
+} as ComponentMeta<typeof ConsumerGroupResetOffset>;
+
+const Template: ComponentStory<typeof ConsumerGroupResetOffset> = (args) => (
+  <ConsumerGroupResetOffset {...args} />
+);
+
+export const ResetOffsetWhenTopicisNotSelected = Template.bind({});
```
jgiardino, line 48: Should "is" have a capital "I" like the others?

jgiardino (commented):
Thanks for making updates! There were a few minor things I noticed regarding accessible labels, but otherwise things look good. 

One question I have is whether the checkboxes in the table should be working as expected? Or is that type of functionality not included in the storybook components?

HemaHG:
sure @riccardo-forina , we will do

HemaHG:
@jenn, selecting the checkbox in the table will be handled in parent component

HemaHG:
@riccardo-forina , @jgiardino I have addressed all the review comment, could you guys please review

src/Kafka/ConsumerGroups/components/TopicSelect.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <Select
+      aira-label={t("consumerGroup.reset_offset_topic_label")}
+      toggleAriaLabel={t("consumerGroup.reset_offset_topic_label")}
+      id={"topic-select"}
+      variant={SelectVariant.single}
+      onToggle={onToggle}
+      onSelect={onSelect}
+      isOpen={isOpen}
+      selections={value}
+      placeholderText={t("common:select")}
```
riccardo-forina, line 41: Is this the right placeholder? "Select" doesn't seem that informative to me.

riccardo-forina: The same applies to all other cases where `common:select` is used.

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,274 @@
...
+      "consumerGroup.member_id"
+    )}`,
+    current_offset: t("consumerGroup.current_offset"),
+    log_end_offset: t("consumerGroup.log_end_offset"),
+    offset_lag: t("consumerGroup.offset_lag"),
+    new_offset: t("consumerGroup.new_offset"),
+  };
+
+  return (
+    <Modal
```
riccardo-forina, line 104: I'd suggest top-aligning the modal. We noticed in the create Kafka one that if the modal's content is dynamic if it's not top-aligned, it moves on the screen. It's disorientating.

src/Kafka/ConsumerGroups/components/TopicSelect.tsx
HemaHG: As per the design doc https://docs.google.com/presentation/d/1wAsmKnco9-ZzJadp1djCim_RqlBgWEumFi2Fbxd531Q/edit#slide=id.gdfee5733b1_6_1, the placeholder text is select

HemaHG: @riccardo-forina, I have added the placeholder test as per the design https://docs.google.com/presentation/d/1wAsmKnco9-ZzJadp1djCim_RqlBgWEumFi2Fbxd531Q/edit#slide=id.gdfee5733b1_6_1. other than that I have address all other review comments

jgiardino (approved):
You updates look good.

There are a couple of open questions related to this, but we can address them in a later PR:
- Should the placeholder text just say "Select"?
- What is the proper implementation of aria- attributes on a Select component so that the selected value is communicated via assistive technology? (posted on [PF slack](https://patternfly.slack.com/archives/C916H09K4/p1659360678163829))

jgiardino:
BTW, I got feedback on the open question about the "Select" placeholder sooner than I expected:

> The only recommendation we have around that is what you shared, to help folks understand when to use "select" vs. when to use "click."
> As placeholder text, I've almost always seen a descriptive word following "Select." For example, "Select topic," as you mention. I see that as being extra clear in that it:
> 
> - Identifies what the user is selecting.
> - Communicates the quantity of topics the user can select (vs. "Select topicS").
> 
> That is a long-winded answer :slightly_smiling_face: So although we might not have specific guidance for this (as I'm aware of), I think it's a good idea to have a specific word after "Select."

src/Kafka/ConsumerGroups/components/OffsetSelect.tsx
```diff
@@ -49,7 +49,7 @@ export const OffsetSelect: FunctionComponent<OffsetSelectProps> = ({
       onSelect={onSelect}
       variant={SelectVariant.single}
       isOpen={isOpen}
-      placeholderText={t("common:select")}
+      placeholderText={t("consumerGroup.select_offset")}
```
HemaHG, line 52: @jenn added a specific word after the "select" text

src/Kafka/ConsumerGroups/components/TopicSelect.tsx
```diff
@@ -38,7 +38,7 @@ export const TopicSelect: FunctionComponent<TopicSelectProps> = ({
       onSelect={onSelect}
       isOpen={isOpen}
       selections={value}
-      placeholderText={t("common:select")}
+      placeholderText={t("consumerGroup.select_topic")}
```
HemaHG, line 41: @Jenn added a specific word after the "select" text

src/Kafka/ConsumerGroups/components/ConsumerGroupResetOffset.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  const isResetOffsetDisabled = (): boolean => {
+    return (
+      !selectedTopic ||
+      !selectedOffset ||
+      !confirmCheckboxChecked ||
+      !isDisconnected ||
+      consumers.filter(({ selected }) => selected === true).length === 0
+    );
+  };
```
riccardo-forina, line 90:
```suggestion
  const isResetOffsetDisabled = 
      !selectedTopic ||
      !selectedOffset ||
      !confirmCheckboxChecked ||
      !isDisconnected ||
      consumers.filter(({ selected }) => selected === true).length === 0;
```

No need for that to be a function.

riccardo-forina (approved):

PR #291, opened 2022-04-12
feat: add ResponsiveTable, TableView, KafkaMessageBrowser

riccardo-forina:

This adds a few new components.

### KafkaMessageBrowser

A component to visualize messages for a topic of a Kafka instance. Visible under Kafka > Kafka Message Browser

### ResponsiveTable

A component to render a table with responsive columns (columns that hide automatically if the available space isn't enough to show them). Visible under Shared > Table > Responsive Table

### TableView

A TableView with pagination and support for custom toolbars. Visible under Shared > Table > Table View

Partially fixes https://issues.redhat.com/browse/MGDSTRM-7801


simrandhaliw:
Thanks for creating these Stories for Message Broswer, Riccardo! 
I just noticed a few differences in the Storybook and designs:

1. Timestamp items in the table have a different format. 
PF suggests these date and time formats: https://www.patternfly.org/v4/ux-writing/numerics/#date-and-time-formats. So maybe we should use this pattern: 21 January 2019, 9:38:11:123 PM

About the time, Pete wanted time to have milliseconds. He mentioned this in a UX review meeting “Timestamp could have a lot of messages coming in (hundreds per second) and might require including milliseconds (could have 10 that came in during the same second)”

2. Timestamp type label seems missing in the drawer description list. However, this wasn’t something very important. So, if we can’t get this value, we can skip showing this in the drawer.

3. In the ‘Example’ story, the search button in the toolbar looks enabled (but it should be disabled initially). Also, the reload button is disabled, but this should stay enabled always.

simrandhaliw:
Katie (one of the content designers on the UXD team) suggested that:

- “No [name of property]” seems like a safe pattern to follow for the null values in the table. For the styling of this text, we can follow what Jenn suggested i.e. use grey text. 

- Also, “Check for new data" sounds perfect in the context of the initial empty state. Looks like we are already using this for the button in the Storybook, so no changes there.

riccardo-forina:
>     1. Timestamp items in the table have a different format.

Should be fixed now.

>     2. Timestamp type label seems missing in the drawer description list. However, this wasn’t something very important. So, if we can’t get this value, we can skip showing this in the drawer.

Yes I didn't include that because it felt like an internal implementation detail rather than something really useful for end users, but I might be wrong 😓 @wtrocki WDYT?

>     3. In the ‘Example’ story, the search button in the toolbar looks enabled (but it should be disabled initially). Also, the reload button is disabled, but this should stay enabled always.

Search is enabled only if you changed any of the filters in the toolbar. Refresh is always enabled unless you changed a filter in the toolbar, where you have to search first to apply the filters.

riccardo-forina:
> Katie (one of the content designers on the UXD team) suggested that:
> 
>     * “No [name of property]” seems like a safe pattern to follow for the null values in the table. For the styling of this text, we can follow what Jenn suggested i.e. use grey text.
> 
>     * Also, “Check for new data" sounds perfect in the context of the initial empty state. Looks like we are already using this for the button in the Storybook, so no changes there.

Both done 👍

.github/workflows/main.yml
```diff
@@ -40,7 +40,9 @@ jobs:
           node-version: 16.6.2
 
       - name: Build storybook
-        run: npm run build-storybook -- --quiet --loglevel silent
+        run: |
```
riccardo-forina, line 43: This was to make the Storybook step to complete even with failing tests.

.github/workflows/reusable-build.yml
```diff
@@ -31,7 +31,9 @@ jobs:
         with:
           node-version: ${{ inputs.node-version }}
       - name: Run the test suite
-        run: npm run test-storybook:ci
+        run: |
```
riccardo-forina, line 34: This was to make the Storybook step complete even with failing tests.

.storybook/main.ts
```diff
@@ -10,6 +10,7 @@ module.exports = {
     "@storybook/addon-interactions",
     "@storybook/addon-a11y",
     "@storybook/addon-jest",
+    "storybook-xstate-addon/preset",
```
riccardo-forina, line 13: Added back this plugin for easier inspection of the xstate machines in Storybook

src/Kafka/KafkaMessageBrowser/components/DateTimePicker.css
```diff
@@ -0,0 +1,17 @@
+.pf-c-form-control.custom-width {
```
riccardo-forina, line 1: Ah, I need to change these classes

simrandhaliw:
@riccardo-forina, thanks for addressing my previous comments! 

A few more notes about the latest merge:
1. For the text "No [name of property]" we should be using sentence case, which means the N in 'No' would be capital just like we have it. However, the name of the property would be in small case. E.g. No key. 
2. Looking at the [PF guidelines for Date and Time format](https://www.patternfly.org/v4/ux-writing/numerics/#date-and-time-formats), I think I prefer this format the most "21 January 2019, 9:38:11:123 PM" (added the milliseconds since we have that in our case) because there is a comma separator between date and time. @jgiardino and @kaedwards, do you think we can/should use: 

   - US date format i.e. month first. E.g. January 21, 2019, 9:38:11:123 PM
   - month abbreviation E.g. Jan 21, 2019, 9:38:11:123 PM or 21 Jan 2019, 9:38:11:123 PM

3. The grey placeholder text 'Select epoch timestamp' in the second filter group seems truncated. Can we make the field a bit wider to show that entire string?
4. The placeholder for Timestamp should be 'YYYY-MM-DDThh:mm:ss.s'. Also, the time picker dropdown doesn't seem scrollable.

riccardo-forina:
>     1. For the text "No [name of property]" we should be using sentence case, which means the N in 'No' would be capital just like we have it. However, the name of the property would be in small case. E.g. No key.

Done

>     2. Looking at the [PF guidelines for Date and Time format](https://www.patternfly.org/v4/ux-writing/numerics/#date-and-time-formats), I think I prefer this format the most "21 January 2019, 9:38:11:123 PM" (added the milliseconds since we have that in our case) because there is a comma separator between date and time.

Added the comma, but kept the dot as a separator between seconds and milliseconds since that's the standard way to do it AFAIK.

>     3. The grey placeholder text 'Select epoch timestamp' in the second filter group seems truncated. Can we make the field a bit wider to show that entire string?

It looks like the `InputGroup` component doesn't allow specifying sizes. Maybe we can make the dropdown label shorter?

>     4. The placeholder for Timestamp should be 'YYYY-MM-DDThh:mm:ss.s'. Also, the time picker dropdown doesn't seem scrollable.

That's the placeholder I see.
image

Anyhow I felt like the usefulness of that time selector wasn't that great. I learned about the [`TimeComponent`](https://www.patternfly.org/v4/components/time-picker/) and tried to use it in combination with the date picker to give max flexibility to the user and came up with this:
image
I think it's working nice, WDYT?

kaedward:
@simrandhaliw in the [Kafka instance drawer](https://docs.google.com/presentation/d/17-oZp7bPAsGQ8wdhAy62MDjcdYSt5b9HSijo-91LhTg/edit#slide=id.g103ce70f2ba_1_188) under "details", we use this format: "Thursday, January 21, 2019 9:38:11 PM"

If this format would work in this situation, that's what I'd recommend for consistency. But please let me know if I'm missing anything.

simrandhaliw:
> @simrandhaliw in the [Kafka instance drawer](https://docs.google.com/presentation/d/17-oZp7bPAsGQ8wdhAy62MDjcdYSt5b9HSijo-91LhTg/edit#slide=id.g103ce70f2ba_1_188) under "details", we use this format: "Thursday, January 21, 2019 9:38:11 PM"
> 
> If this format would work in this situation, that's what I'd recommend for consistency. But please let me know if I'm missing anything.

Thanks for sharing this link, Katie! In this case, we don't want to show the day in the Timestamp column. We just have the date and time. Right now this is what we have implemented: Mar 15, 2022, 10:11:57.102 AM
We are using comma separators before and after the year. Also, [Riccardo suggested](https://github.com/redhat-developer/app-services-ui-components/pull/291#issuecomment-1104936224) that we use the dot separator between seconds and milliseconds. 

LMK if there are any changes you would recommend.

src/Kafka/KafkaMessageBrowser/components/FilterGroup.tsx
```diff
@@ -0,0 +1,145 @@
...
+            value={timestamp ? timestamp * 1000 : undefined}
+            onChange={onTimestampChange}
+          />
+        )}
+        {currentCategory === "epoch" && (
+          <TextInput
+            isDisabled={isDisabled}
+            type={"number"}
+            aria-label={t("filter.epoch_aria_label")}
+            placeholder={t("filter.epoch_placeholder")}
```
jgiardino, line 132: Playing around with classes currently provided, one way to modify the width of an input within the `<InputGroup>` component is to apply the following attributes and values:
`<TextInput className="pf-u-flex-basis-auto pf-u-flex-grow-0 pf-u-w-initial" size="{length of placeholder text}...`

Since InputGroup is display:flex, we can use the flex utility classes to modify how this element displays so that with basis:auto and grow:0, the input can be the size it wants to be.

Then the `size` attribute can dictate how long the input should be.

If we can specify a size to be based on the length of the placeholder text, that should result in a text input that is long enough to display the placeholder text.

locales/en/message-browser.json
```diff
@@ -0,0 +1,44 @@
...
+    "value": "Value"
+  },
+  "filter": {
+    "epoch": "Epoch timestamp",
+    "epoch_aria_label": "Select epoch timestamp",
+    "epoch_placeholder": "Select epoch timestamp",
+    "latest": "Select latest",
+    "offset": "$t(field.offset)",
+    "offset_aria_label": "Select offset",
+    "offset_placeholder": "Select offset",
```
jgiardino, line 19: One question I had for @simrandhaliw is whether using "Select ..." in the text inputs is appropriate. This is recommended for selecting an option in a menu, but I don't see an equivalent recommendation when the value is manually specified by the user. I'm wondering if instead of "Select epoch timestamp" should be replaced with "Specify epoch timestamp".

Same comment applies here for the offset placeholder.

simrandhaliw: Yes, I agree. Using 'Select' isn't proper in this case. I asked for suggestions in the content-help slack channel, and the two suggestions I received were to use 'Specify' or 'Enter' instead of 'Select'.

src/Kafka/KafkaMessageBrowser/KafkaMessageBrowser.tsx
```diff
@@ -0,0 +1,365 @@
...
+                onClose={deselectMessage}
+              />
+            }
+          >
+            <Toolbar
+              className={"mas-KafkaMessageBrowser-Toolbar"}
+              data-testid={"message-browser-toolbar"}
+            >
+              <ToolbarContent>
+                <ToolbarToggleGroup toggleIcon={<FilterIcon />} breakpoint="xl">
```
jgiardino, line 215: Since this breakpoint is dependent on the viewport size, and since the drawer provides the ability to alter the available area for the toolbar without triggering this responsive behavior, would it work to have this value conditionally set to `2xl` when the drawer is visible to force the toggle to display sooner?

riccardo-forina: 👍

jgiardino:
Thanks @riccardo-forina and @simrandhaliw for working on this. These updates look really good! 

One last thing I noticed is that the Offset range displays in the example that doesn't show data, but doesn't show in the example with data. I remember some discussion around this, but don't remember if there was a decision to change whether we display that. Should it be removed from this view? Should it also be added to the view that shows data?

image

image

riccardo-forina:
> Thanks @riccardo-forina and @simrandhaliw for working on this. These updates look really good!
> 
> One last thing I noticed is that the Offset range displays in the example that doesn't show data, but doesn't show in the example with data. I remember some discussion around this, but don't remember if there was a decision to change whether we display that. Should it be removed from this view? Should it also be added to the view that shows data?
> 
> image
> 
> image

Done :)

jgiardino (approved):
Thanks for making updates! LGTM

PR #304, opened 2022-04-21
chore: Update the banner from Development preview to Service preview

indraraj:

Update the text on the banner from "Development preview" to "Service preview"

fixes # https://issues.redhat.com/browse/MGDCTRS-583 

**Verification steps** 

image

http://localhost:6006/?path=/story/shared-development-preview-development-preview--show-dev-preview


jgiardino (approved):
Your UI updates look good! 

The only suggestion I have is to update the label for this UI component in the storybook nav so that it matches the text change in the example:
image

riccardo-forina:
That would require changing the component name. I don't disagree but it's a breaking change that will require human intervention on app-services-ui

PR #327, opened 2022-05-02
update kafka details tab component

ajaypratap003:

> update kafka details tab component

> https://issues.redhat.com/browse/MGDSTRM-7805

**Verification steps** 


wtrocki:
CC @jgiardino

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.stories.tsx
```diff
@@ -40,22 +40,19 @@ TrialInstanceJustCreated.args = {
   expiryDate: addHours(new Date(), 48),
   instanceType: "eval",
```
wtrocki, line 41: Unrelated to PR. Do you know what this does?

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -63,61 +58,59 @@ export const KafkaDetailsTab: FunctionComponent<KafkaDetailsTabProps> = ({
 
   return (
     <div className="mas--details__drawer--tab-content">
-      <DetailsTabAlert expiryDate={expiryDate} />
+      {instanceType !== "standard" && expiryDate && (
```
wtrocki, line 61: I like how @riccardo-forina done creation screen setting up some states as input logic.
Maybe it could work here as well.

States like: hasExpiration etc.

wtrocki:
If we use x1 we do not need to mention streaming unit. X1 is size id. Streaming unit will be quota consumed value

jgiardino:
Regarding this comment from @wtrocki:
>  If we use x1 we do not need to mention streaming unit. X1 is size id. Streaming unit will be quota consumed value

The expected UI _is_ to display the number of streaming units, and not the size id. 

There is a bullet point in https://architecture.appservices.tech/adr/81/ that the size `id` is not user-facing and is an internal id only. I think @wtrocki is right that `quotaConsumed` is the correct value when displaying the number of streaming units for an instance. I also saw that @riccardo-forina sent an email to confirm that `quotaConsumed` is the correct value, so hopefully that will be confirmed soon.

jgiardino:
Thanks for making these updates, @ajaypratap003! 

Aside from displaying the number of streaming units, the only thing that I see that needs to be updated is outside the scope of this PR. Originally we displayed these values together, and that's why the units included "each" at the end. But with "Ingress" and "Egress" displaying as separate values, we no longer need the "each". 

image

@riccardo-forina or @ajaypratap003 Can one of you open a separate PR where we store text strings to make that edit? I'm seeing this in both the details that display in the Kafka Instance Drawer and also in the [Instance Info](https://619e1921b3e38b003afeb2be-tvvapjvikl.chromatic.com/?path=/story/kafka-create-kafka-instance-with-sizes-components-instance-info--quota-instance) component for the new Create modal.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.stories.tsx
ajaypratap003: expiryDate and instanceType are  using to show expiration message on details page for trial/eval kafka.
for example: `This trial instance will expire in 15 hours 34 minutes`

ajaypratap003:
> Regarding this comment from @wtrocki:
> 
> > If we use x1 we do not need to mention streaming unit. X1 is size id. Streaming unit will be quota consumed value
> 
> The expected UI _is_ to display the number of streaming units, and not the size id.
> 
> There is a bullet point in https://architecture.appservices.tech/adr/81/ that the size `id` is not user-facing and is an internal id only. I think @wtrocki is right that `quotaConsumed` is the correct value when displaying the number of streaming units for an instance. I also saw that @riccardo-forina sent an email to confirm that `quotaConsumed` is the correct value, so hopefully that will be confirmed soon.

Agreed, we need to sow numeric value but we are not getting it from backend in list of kafkas api. 
We are not getting  `quotaConsumed` in list of kafkas api. We are getting size_id from backend.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
ajaypratap003: I'm not aware about hasExpiration. Can you share link where we are using hasExpiration ?

riccardo-forina:
Chiming in real quick just to say to make this a V2 component as @HemaHG  did for the Overview and Kafka pushed components. So only addictions and no changes to existing files other than the index.ts ones

ajaypratap003:
> Thanks for making these updates, @ajaypratap003!
> 
> Aside from displaying the number of streaming units, the only thing that I see that needs to be updated is outside the scope of this PR. Originally we displayed these values together, and that's why the units included "each" at the end. But with "Ingress" and "Egress" displaying as separate values, we no longer need the "each".
> 
> image
> 
> @riccardo-forina or @ajaypratap003 Can one of you open a separate PR where we store text strings to make that edit? I'm seeing this in both the details that display in the Kafka Instance Drawer and also in the [Instance Info](https://619e1921b3e38b003afeb2be-tvvapjvikl.chromatic.com/?path=/story/kafka-create-kafka-instance-with-sizes-components-instance-info--quota-instance) component for the new Create modal.

@jgiardino Opened a PR for translation update https://github.com/redhat-developer/app-services-ui-components/pull/332

ajaypratap003:
> Chiming in real quick just to say to make this a V2 component as @HemaHG did for the Overview and Kafka pushed components. So only addictions and no changes to existing files other than the index.ts ones

This component is not used in kas-ui, kas-ui has own kafkaDetails component. This is unused component in kas-ui. I think we don't need v2 component for this. 

As per discussion with @wtrocki he want to move new kafka details changes on stage and prod. So that user can see kafka storage details on kafka details page. 
I checked backend team has deployed large kafka changes on prod as well. 
@wtrocki wdyt ? should we move this component on prod so user can see storage details. As you said that user can create large kafka through CLI.

riccardo-forina:
Then remove the streaming components part from this, and we’ll make a V2 with that when we’ll be able to show it

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -63,61 +58,59 @@ export const KafkaDetailsTab: FunctionComponent<KafkaDetailsTabProps> = ({
...
-                })
-              )}
-              {renderTextListItem(
-                t("message_size"),
-                t("create-kafka-instance:message_size_value", {
-                  value: messageSize,
-                })
-              )}
-            </>
+          {instanceType === "standard" &&
```
riccardo-forina, line 66: Ideally, the namespace for this shouldn't be `create-kafka-instance` but 'kafka-details` or something like that, with its own json.

Since it's a bit of a pain, it's ok to reuse an existing namespace but please use the new `create-kafka-instance-with-size` which is the one created to support large Kafka.

riccardo-forina (changes requested):

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
```diff
@@ -20,17 +19,14 @@ type KafkaDetailsTabProps = {
...
   region: string;
   instanceType: InstanceType;
-  size?: number;
-  ingress: number;
-  egress: number;
-  storage: number;
+  size?: string;
+  ingress: string;
+  egress: string;
+  storage: string;
```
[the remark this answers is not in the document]
ajaypratap003: API is returning string in response. Even some properties includes unit with value. 
I'll make these types number.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
ajaypratap003: yes, will use new component `create-kafka-instance-with-size`

locales/en/create-kafka-instance.json
```diff
@@ -16,6 +16,7 @@
   "create_instance_name_helper_text": "Must start with a letter and end with a letter or number. Valid characters include lowercase letters from a to z, numbers from 0 to 9, and hyphens ( - ).",
   "create_instance_name_helper_text_name_taken": "{{name}} already exists. Try a different name.",
   "duration_value": "{{value}} hours",
+  "egress_value": "up to {{value}} MiB/second",
```
riccardo-forina, line 19: Can you restore this file as it was in the main branch, please?

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.stories.tsx
```diff
@@ -30,32 +30,28 @@ const instanceInfo = {
 export const StandardInstanceCreated = Template.bind({});
 StandardInstanceCreated.args = {
   instanceType: "standard",
-  size: 1,
+  size: "x1",
```
riccardo-forina, line 33:
```suggestion
  size: "1",
```

We should not hint that size matches the streaming unit id, so let's use a generic numeric "name" for it.

src/Kafka/KafkaInstanceDrawer/KafkaDetailsTab.tsx
riccardo-forina:
```suggestion
      {expiryDate && (
```

I'd keep the logic tied to the presence of the `expiryDate`. Checking for the instance type doesn't add much value imho.

riccardo-forina (approved):
I left a few comments that would be nice to address.

PR #333, opened 2022-05-04
fix: changed limits for kafka instance

wtrocki:
Does change limits for kafka instances.

### DIFF
--- a/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
+++ b/src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
@@ -65,12 +65,12 @@ export const CreateKafkaInstance: FunctionComponent<
   onCreate,
   disableFocusTrap,
   trialDurationInHours = 48,
-  ingress = 1,
-  egress = 1,
-  storage = 10,
-  maxPartitions = 100,
-  connections = 100,
-  connectionRate = 50,
+  ingress = 50,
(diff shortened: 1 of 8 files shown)


src/Kafka/CreateKafkaInstance/CreateKafkaInstance.tsx
```diff
@@ -119,6 +119,15 @@ export const CreateKafkaInstance: FunctionComponent<
     [create]
   );
 
+  if (!isTrial) {
+    ingresEgress = 100;
```
wtrocki, line 123: For ingressEgress there are two values in doc 100 and 50.

wtrocki: This is for standard instance

jgiardino: You're correct. The updates we made for [the new UI](https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-create-kafka-instance-with-sizes-components-instance-info--quota-instance) split these out into separate values. I'm not sure what our options are for splitting them out for this PR. Maybe this is something that @riccardo-forina can provide a suggestion for.

ajaypratap003 (approved):
LGTM

wtrocki:
@riccardo-forina will take this over to improve component messaging.

PR #337, opened 2022-05-09
chore: Update OUIA-ID tags for pricing plan

HemaHG:

> Update OUIA-ID tags for pricing plan in kafka overview page

https://issues.redhat.com/browse/MGDSTRM-7927

### DIFF
--- a/src/AppServices/KafkaPage/KafkaPageV2.tsx
+++ b/src/AppServices/KafkaPage/KafkaPageV2.tsx
@@ -46,7 +46,7 @@ export const KafkaPageV2: FunctionComponent = () => {
       />
       <MarketingPageSection>
         <Grid hasGutter lg={6}>
-          <Card>
+          <Card ouiaId="card-overview-purchase-now">
             <CardHeader>
               <CardHeaderMain>
                 <CardTitle>
@@ -64,13 +64,14 @@ export const KafkaPageV2: FunctionComponent = () => {
                 component="a"
                 href="https://marketplace.redhat.com/en-us/products/red-hat-openshift-streams-for-apache-kafka"
                 target="_blank"
+                ouiaId="link-purchase-now"
               >
                 {t("purchaseCardCallToActionButton")}
                 <ExternalLinkAltIcon className="pf-u-ml-md" />
               </Button>
             </CardFooter>
           </Card>
-          <Card>
+          <Card ouiaId="card-overview-contact-sales">
             <CardHeader>
               <CardTitle>
                 <Title headingLevel="h2">{t("contactSalesCardTitle")}</Title>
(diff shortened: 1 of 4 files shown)


src/AppServices/KafkaPage/KafkaPageV2.tsx
```diff
@@ -64,13 +64,14 @@ export const KafkaPageV2: FunctionComponent = () => {
                 component="a"
                 href="https://marketplace.redhat.com/en-us/products/red-hat-openshift-streams-for-apache-kafka"
                 target="_blank"
+                ouiaId="link-purchase-now"
```
HemaHG, line 67: The Jira task suggested adding data-ouia-id to the link, but I am really not sure the link refers to the button component

riccardo-forina:
Sorry for the late review. I forgot to say when assigning this to you that we should [add unit tests around any ouia id](https://github.com/redhat-developer/app-services-ui-components/blob/9422674d3382888a0dcfe9b4a446086f3fd15156/src/Kafka/CreateKafkaInstanceWithSizes/CreateKafkaInstanceWithSizes.test.tsx) we add, to be certain that we don't change/drop them without us noticing. Snapshot-based testing isn't enough since it's too easy to miss a change to the id, especially if you don't know how important it is not to change them.

src/AppServices/KafkaPage/KafkaPage.test.tsx
```diff
@@ -10,4 +10,30 @@ describe("KafkaPage", () => {
...
+    await waitForI18n(comp);
+
+    const purchaseCard = comp.getByRole("article", { name: "Contact sales" });
+    expect(purchaseCard.dataset.ouiaComponentId).toBe(
+      "card-overview-contact-sales"
+    );
+
+    const link = comp.getByRole("link", { name: "Contact sales" });
+    expect(link.dataset.ouiaComponentId).toBe("link-contact-sales");
+  });
```
riccardo-forina, line 38:
```suggestion
  it("should persist ouia ids", async () => {
    const comp = render(<Example />);
    await waitForI18n(comp);

    expect(comp.getByRole("article", { name: "Purchase now" }))
      .toHaveAttribute("data-ouia-component-id", "card-overview-purchase-now");
    expect(comp.getByRole("link", { name: "Purchase now" }))
      .toHaveAttribute("data-ouia-component-id", "link-purchase-now");
  
    expect(comp.getByRole("article", { name: "Contact sales" }))
      .toHaveAttribute("data-ouia-component-id", "card-overview-contact-sales");
    expect(comp.getByRole("link", { name: "Contact sales" }))
      .toHaveAttribute("data-ouia-component-id", "link-purchase-now");
  });
```

I haven't tested this, but I'd simplify the test like this. It's ok to test more than one thing in a single test. To test attributes there is a specific `toHaveAttribute` function you can use.

riccardo-forina (approved):

PR #340, opened 2022-05-11
Large kafka updates

riccardo-forina:

Renders the size slider using a `displayName` value for the label. It still uses the quota under the hood to calculate if the selected size goes over quota, even if we still display it as a "streaming unit" to the user.

Also, add a "Tech preview" alert to the Size field for sizes flagged as "preview".
To support this alert, this adds a new `ExternalLink` shared component that eases working with the `Trans` component.

https://issues.redhat.com/browse/MGDSTRM-8432

**Verification steps** 

1. Kafka > Create Kafka Instance With Sizes > Create Kafka Instance With Sizes: this shows the "Tech preview" for the second size, and the error message for over quota without the "Tech preview" alert for the third size.
2. Shared > External link > External link: this is the new component mentioned above

### DIFF
--- a/src/shared/ExternalLink/ExternalLink.tsx
+++ b/src/shared/ExternalLink/ExternalLink.tsx
@@ -1,0 +1,28 @@
+import { Button, ButtonProps, ButtonVariant } from "@patternfly/react-core";
+import { ExternalLinkAltIcon } from "@patternfly/react-icons/dist/js/icons/external-link-alt-icon";
+import { FunctionComponent } from "react";
+
+export type ExternaLinkProps = {
+  testId: string;
+  target?: ButtonProps["target"];
+  href: NonNullable<ButtonProps["href"]>;
+};
+
+export const ExternalLink: FunctionComponent<ExternaLinkProps> = ({
+  testId,
+  target = "_blank",
+  href,
+  children,
+}) => (
+  <Button
+    data-testid={testId}
+    isInline
+    variant={ButtonVariant.link}
+    component="a"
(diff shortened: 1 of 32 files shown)


src/shared/ExternalLink/ExternalLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+  <Button
+    data-testid={testId}
+    isInline
+    variant={ButtonVariant.link}
+    component="a"
+    target={target}
+    href={href}
+  >
+    {children}
+    <ExternalLinkAltIcon className="pf-u-ml-xs" />
```
jgiardino, line 26: I really like that you created a shared component for this. Something I noticed when we were using this pattern for the Overview page was that depending on the size of the viewport, the icon might wrap to the next line. This could be potentially confusing where it's less obvious that the link icon is meant to describe the behavior of the link, because they're not adjacent in that case.

Below is an example of this component showing that the icon wraps. I realize this is an edge case, and am not sure if there's a way to solve that (e.g. can we use `&nbsp` between the icon and text here?), but wanted to just mention this in case it's an easy thing to address. 

image

jgiardino (approved):
I had one comment about something that would be nice to address if possible, but otherwise these updates look great!

src/shared/ExternalLink/ExternalLink.tsx
riccardo-forina: Nice catch! I think this should not happen anymore now :)

PR #344, opened 2022-05-12
feat: show dates in the UTC tz

riccardo-forina:

Dates are confusing. We found out that the consoledot guidelines recommend not localizing dates and showing them in the UTC timezone. The same ask has been made while demoing the message browser.

This PR changes the default behaviour of the `FormatDate` when used with either the `long` or `longWithMilliseconds` format to show the date in the UTC tz.

**Verification steps** 

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,6 +32,11 @@
           {
             "name": "dayjs",
             "message": "Please use date-fns instead"
+          },
+          {
+            "name": "date-fns",
+            "importNames": ["format"],
+            "message": "Please use format or formatInTimeZone from date-fns-tz instead"
           }
         ]
       }
--- a/locales/en/metrics.json
+++ b/locales/en/metrics.json
@@ -1,8 +1,8 @@
 {
   "all_topics": "All topics",
   "axis-label-bytes": "Bytes",
-  "axis-label-time": "Time",
-  "axis-label-time-full": "Time (HH:mm MM/dd)",
+  "axis-label-time": "UTC Time",
+  "axis-label-time-full": "UTC Time (HH:mm MM/dd)",
   "chart-popover-icon-screenreader-text": "Information about {{title}}",
   "client_connections": "Client connections",
   "client_connections_helper_text": "Client connections are the total connections across all clients over time. Clients can make multiple connections to multiple brokers in the Kafka instance. This metric enables you to review client activity in the instance and to assess available connections relative to the limit.",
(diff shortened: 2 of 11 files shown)


jgiardino (approved):
Thanks for making this update! No changes from me, but I'm questioning how we reflect this in the metrics page. Should the x-axis label be "Time (UTC)"?

riccardo-forina:
> Thanks for making this update! No changes from me, but I'm questioning how we reflect this in the metrics page. Should the x-axis label be "Time (UTC)"?

Good catch. I went with `UTC Time` to support the 2 different labels we have:

```
  "axis-label-time": "UTC Time",
  "axis-label-time-full": "UTC Time (HH:mm MM/dd)",
```

WDYT?

jgiardino:
@riccardo-forina - Thanks for mentioning that! I completely forgot we had two labels. 

I was chatting with @kaedward about this. In your suggestion, it's a little odd to not have "Time" as the primary / prominent text for the label. But we don't really have a great suggestion for this other than the following:

> Time (UTC, HH:mm MM/dd)
> Time (UTC)

Looking across other examples of charts in consoledot, there are several cases where month is spelled out with 3 characters, as shown in this example of the [sub watch page for Kafka](https://marvelapp.com/prototype/8b97d9f/screen/82017465). Maybe this doesn't work for localization, but we wanted to throw this out in case it's an option. Using the 3 char format for month lets us not display the `HH:mm MM/dd` format in the label. 

- Time & date use this format for the x-axis tickmark label:
  > 12:30
  > May 3
- And the x-axis label always displays as this:
  > Time (UTC)

riccardo-forina:
> @riccardo-forina - Thanks for mentioning that! I completely forgot we had two labels.
> 
> I was chatting with @kaedward about this. In your suggestion, it's a little odd to not have "Time" as the primary / prominent text for the label. But we don't really have a great suggestion for this other than the following:
> 
> > Time (UTC, HH:mm MM/dd)
> > Time (UTC)
> 
> Looking across other examples of charts in consoledot, there are several cases where month is spelled out with 3 characters, as shown in this example of the [sub watch page for Kafka](https://marvelapp.com/prototype/8b97d9f/screen/82017465). Maybe this doesn't work for localization, but we wanted to throw this out in case it's an option. Using the 3 char format for month lets us not display the `HH:mm MM/dd` format in the label.
> 
>     * Time & date use this format for the x-axis tickmark label:
>       > 12:30
>       > May 3
> 
>     * And the x-axis label always displays as this:
>       > Time (UTC)

Great feedback, thanks! I removed the extended x-axis label since tick labels now show the date in the MMM format (so, 3 letters per month)

PR #345, opened 2022-05-13
fix: support single sizes

riccardo-forina:

Integrating the create Kafka instance with sizes with the new APIs, I realized that trial instances come with a single size. The slider wasn't working properly with a single size, this adds an extra scenario for that where the slider is rendered with a faux 0 size that can't be selected.

**Verification steps** 

Kafka > Create Kafka Instance With Size > Components > Field Size

image

### DIFF
--- a/src/Kafka/CreateKafkaInstanceWithSizes/components/FieldSize.stories.tsx
+++ b/src/Kafka/CreateKafkaInstanceWithSizes/components/FieldSize.stories.tsx
@@ -81,7 +81,14 @@ MvpOverQuota.args = {
 export const Trial = Template.bind({});
 Trial.args = {
   value: 1,
-  sizes: summitSizes,
+  sizes: [{ id: "x1", quota: 1, displayName: "1", status: "stable" }],
   remainingQuota: 1,
   validity: "trial",
+};
+
+export const SingleSize = Template.bind({});
+SingleSize.args = {
+  value: 1,
+  sizes: [{ id: "x1", quota: 1, displayName: "1", status: "stable" }],
+  remainingQuota: 1,
 };
--- a/src/Kafka/CreateKafkaInstanceWithSizes/components/FieldSize.tsx
+++ b/src/Kafka/CreateKafkaInstanceWithSizes/components/FieldSize.tsx
@@ -68,14 +68,18 @@ export const FieldSize: VoidFunctionComponent<FieldSizeProps> = ({
       />
     );
   }
+  const atLeastTwoSizes =
+    sizes.length > 1 ? sizes : [{ id: "", quota: "" }, ...sizes];
 
-  const valueIndex =
-    validity !== "trial" ? sizes.findIndex((size) => size.quota === value) : -1;
-
-  const steps: SliderProps["customSteps"] = sizes.map((s, index) => ({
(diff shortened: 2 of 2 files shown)


jgiardino:
This might be a bad design, but the [original design](https://docs.google.com/presentation/d/1U4j27K3L-9Ry0GPe-fcKoU48dCPlEJsfoXJUpldAY3U/edit#slide=id.g118b16caf36_0_0) was to accommodate this use case:
> If I don't have a subscription and can only create a trial instance, then I can get some idea about what options are available with a subscription. 

In this use case, the size option isn't meant to show me what trial sizes are available, but rather what sizes are available with a subscription. This is why in the example above, we show that there are two options, 1 SU and 2 SU, along with a message that size options are only available when you have a subscription. 

I realize that how we're trying to convey these options in the UI isn't a clean map to what the API gives us, and it would be good to refactor the UI to better reflect what we're getting from the API. But when we chatted with Bilgin last week, it sounded like we likely have half a year to refactor this UI. I also would like to better understand what the next evolution of trial instances are and any expectations regarding how we represent trial instances to the user before refactoring this UI.

For this upcoming release, I would suggest we keep what's currently implemented in [Quota Available - Trial](https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-create-kafka-instance-with-sizes-stories-form-load--trial-available-on-form-load), since this matches the design as expected. 

But also, we definitely should not use the concept of a streaming unit when describing a trial instance.  When we do refactor the UI, we need to remove the mapping of sizes to a "streaming unit" and just display the `displayName` property of a size, and maybe not use a slider at all.

riccardo-forina:
You are right, the way I implemented this dialog doesn't really support showing the standard sizes while creating an instance. The story you linked doesn't match how the system actually works. I'll make a new PR to accommodate this feedback.

Still, I think we should consider the case where we are getting a single size. The Slider component doesn't really work well with a single value, hence I opened this to fake a zero, nonselectable value just to have the slider adequately rendered.

jgiardino:
> The Slider component doesn't really work well with a single value, hence I opened this to fake a zero, nonselectable value just to have the slider adequately rendered.

💯 Also, the slider doesn't really work well with how sizes are defined in the API. Sliders expect numeric values. Even if we can display a non-numeric value as labels on the slider, there is still an expectation that the tick marks are positioned according to the numeric value associated with that non-numeric label.  

> The story you linked doesn't match how the system actually works.

I agree, but there are other problems with the current UI design besides that, which should also be addressed. 

Following our discussion last week with Bilgin, my take-away was that we keep the UI design as-is for now , because we know the data that the API will provide for at least the rest of the year will remain as-is (i.e. standard type with 2 sizes, along with a dev type trial instance). And we use that time to refactor the UI to better reflect the underlying system. 

I started to organized notes in this Miro board to capture what updates to the UI we should make now vs longer term. This frame captures my thoughts about longer-term updates we should make, but there are likely other considerations to make related to the dev/trial instances: 
https://miro.com/app/board/uXjVO3wbhfM=/?moveToWidget=3458764524754870676&cot=14

> Still, I think we should consider the case where we are getting a single size.

Is this something that you think can be addressed as part of a future refactor of the UI? Or are you thinking this is needed for our release this week?

riccardo-forina:
We can close this, I opened another PR that actually implements the design we have with respect to how the system work.

PR #350, opened 2022-05-17
feat: support different sizes for standard and trial instances in the create Kafka dialog with sizes

riccardo-forina:

[This addresses @jgiardino comment here:](https://github.com/redhat-developer/app-services-ui-components/pull/345#issuecomment-1128198906)

> This might be a bad design, but the [original design](https://docs.google.com/presentation/d/1U4j27K3L-9Ry0GPe-fcKoU48dCPlEJsfoXJUpldAY3U/edit#slide=id.g118b16caf36_0_0) was to accommodate this use case:
> 
> > If I don't have a subscription and can only create a trial instance, then I can get some idea about what options are available with a subscription.
> 
> In this use case, the size option isn't meant to show me what trial sizes are available, but rather what sizes are available with a subscription. This is why in the example above, we show that there are two options, 1 SU and 2 SU, along with a message that size options are only available when you have a subscription.
> 
> I realize that how we're trying to convey these options in the UI isn't a clean map to what the API gives us, and it would be good to refactor the UI to better reflect what we're getting from the API. But when we chatted with Bilgin last week, it sounded like we likely have half a year to refactor this UI. I also would like to better understand what the next evolution of trial instances are and any expectations regarding how we represent trial instances to the user before refactoring this UI.
> 
> For this upcoming release, I would suggest we keep what's currently implemented in [Quota Available - Trial](https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-create-kafka-instance-with-sizes-stories-form-load--trial-available-on-form-load), since this matches the design as expected.
> 
> But also, we definitely should not use the concept of a streaming unit when describing a trial instance. When we do refactor the UI, we need to remove the mapping of sizes to a "streaming unit" and just display the `displayName` property of a size, and maybe not use a slider at all.

**Verification steps** 

Trial instances show the sizes available for standard instances in the slider, but the limits on the right pane are those for the trial instance.
image

Standard instances still work as they should
image

### DIFF
--- a/src/Kafka/CreateKafkaInstanceWithSizes/Stories/storiesHelpers.tsx
+++ b/src/Kafka/CreateKafkaInstanceWithSizes/Stories/storiesHelpers.tsx
@@ -18,7 +18,6 @@ import {
   Provider,
   ProviderInfo,
   Providers,
-  Size,
 } from "../types";
 
 export const AWS: ProviderInfo = {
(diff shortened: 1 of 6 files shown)


src/Kafka/CreateKafkaInstanceWithSizes/Stories/storiesHelpers.tsx
```diff
@@ -50,76 +49,113 @@ export const AZURE: ProviderInfo = {
...
+        connections: 60,
+        connectionRate: 70,
+        maxPartitions: 80,
+        messageSize: 90,
+        trialDurationHours: undefined,
+      },
+      {
+        id: "x3",
+        displayName: "3",
+        quota: 5,
```
jgiardino, line 86: I'm not sure I fully understand how this data relates to the implementation in production, so this comment might be irrelevant.

One of the updates we discussed for the release we want to make this week is to use `displayName` for the option label on the slider, but still use `quotaConsumed` for defining the numeric value of that option.

src/Kafka/CreateKafkaInstanceWithSizes/Stories/storiesHelpers.tsx
```diff
@@ -50,76 +49,113 @@ export const AZURE: ProviderInfo = {
...
+        connectionRate: 7,
+        maxPartitions: 8,
+        messageSize: 9,
+        trialDurationHours: undefined,
+      },
+      {
+        id: "x2",
+        displayName: "2",
+        quota: 2,
+        status: "preview",
```
jgiardino, line 73: Similar to my other comment, I'm not sure how this data relates to how we will use this component in production, but looking at the [current implementation](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/managed-services/cicd/saas/saas-kas-fleet-manager.yaml#L311), this value is provided as `maturityStatus`.

jgiardino (approved):
I don't fully understand the updates you made in this PR, but looking at the UI preview, everything looks as expected. 😄  I made a few inline comments about values we expect from the API, but feel free to ignore.

riccardo-forina:
The answer to both comments is that the UI defined its own data model which might differ from the one used by the API. We do that to enforce stricter types and to [decouple](https://en.m.wikipedia.org/wiki/Law_of_Demeter) the UI from the API.

PR #351, opened 2022-05-18
fix: trial instances don't check for the correctness of the selected size since it can't be selected

riccardo-forina:

Fixes a bug where the form would show an error alert asking to fix form errors when creating trial instances and with 0 as the available quota.

### DIFF
--- a/src/Kafka/CreateKafkaInstanceWithSizes/CreateKafkaInstanceMachine.ts
+++ b/src/Kafka/CreateKafkaInstanceWithSizes/CreateKafkaInstanceMachine.ts
@@ -508,10 +508,14 @@ const CreateKafkaInstanceMachine = createMachine(
       noSizes: ({ sizes }) => sizes === undefined,
       emptySizes: ({ sizes }) =>
         sizes !== undefined && sizes.standard.length === 0,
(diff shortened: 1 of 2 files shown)


ajaypratap003 (approved):
LGTM

src/Kafka/CreateKafkaInstanceWithSizes/CreateKafkaInstanceMachine.ts
```diff
@@ -508,10 +508,14 @@ const CreateKafkaInstanceMachine = createMachine(
       noSizes: ({ sizes }) => sizes === undefined,
       emptySizes: ({ sizes }) =>
         sizes !== undefined && sizes.standard.length === 0,
-      sizeIsInQuota: ({ form, capabilities }) =>
-        capabilities !== undefined &&
-        form.size !== undefined &&
-        form.size.quota <= capabilities.remainingQuota,
+      sizeIsInQuota: ({ form, capabilities }) => {
+        if (capabilities === undefined) return false;
+        if (capabilities.plan === "trial") return true;
```
wtrocki, line 513: app-sevices-ui-shared has plan constants defined. Can we use that module in this module?

PR #373, opened 2022-05-25
fix: use trial duration from the API

ajaypratap003:
Add dynamic duration (lifespan_seconds) instead hard code 48 hours

https://issues.redhat.com/browse/MGDSTRM-7914

**Verification steps** 

- User trial kafka and open create kafka modal 
- There is a expiration time duration which is based on api field `lifespan_seconds`

image

Rest part adding duration in table that will be covered in kas-ui in separate PR.
https://github.com/bf2fc6cc711aee1a0c2a/kas-ui/pull/1181

### DIFF
--- a/locales/en/create-kafka-instance-with-sizes.json
+++ b/locales/en/create-kafka-instance-with-sizes.json
@@ -67,7 +67,7 @@
     "regions_unavailable_message": "Cloud provider regions are temporarily unavailable. Try again later.",
     "system_unavailable_message": "$t(common:unexpected_error)",
     "system_unavailable_title": "$t(common:something_went_wrong)",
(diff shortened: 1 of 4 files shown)


jgiardino:
@riccardo-forina pointed out how to access the preview. I was able to look at the UI examples and didn't see any issues.

riccardo-forina:
@ajaypratap003 could you please fill out the template with the info we ask to provide? Leaving the defaults isn't that nice.

ajaypratap003:
> The storybook build appears to be not working lately. Without being able to see a preview, I think this PR updates the create modal to display the duration based on what information is provided by the API. If so, that all seems good.
> 
> As for the original jira, there should also be updates to the table view. But these are probably needed in the kas-ui repo. Just wanted to confirm that those updates were also in progress.

@jgiardino yes, this change for alert and table duration change will be in kas-ui

riccardo-forina (approved):

PR #374, opened 2022-05-25
fix: improve create Kafka dialog UX

riccardo-forina:

This fixes a number of issues with the create Kafka dialog.

### Sizes can be disabled

An instance size could be visible but not selectable for instance creation. It shows with a warning.
image

Trying to submit the form with a disabled size selected will transform the warning in an error.
image

### Generic alert when trial instances are unavailable for technical reasons

Always display the "All available trial instances are currently in use. Try again later." message for all cases where trial instances can not be created for reasons unrelated to the user.

### Top aligned modal

Aligning the modal to the top of the screen reduces the amount of layout jumping when interacting with elements that trigger changes on the visualized UI, like changing provider/region where skeletons are displayed while loading the new data.

Skeletons were tweaked to better reflect the space taken by the real elements, again to reduce layout jumping.

### State machine refactoring

Refactor the state machine to rely less on the context and more on its states. This will improve the future maintainability of the machine.

### Fix the XState inspector height issue

The inspector will use the available height instead of a fixed 1000px.

### Improve rollup.js build

Suppress the warnings, by properly acknowledging them. Change the config so that future warnings are promoted to errors, forcing builds to fail.

### DIFF
--- a/locales/en/metrics.json
+++ b/locales/en/metrics.json
@@ -34,6 +34,8 @@
   "metric_kpi_topics_description": "Topics are event logs in a $t(common:kafka) instance. This metric shows the total number of topics in the $t(common:kafka) instance.",
   "metric_kpi_topics_name": "Topics",
   "metric_not_available": "Data unavailable",
+  "metrics_lag_description": "Metrics regularly experience lag, and do not automatically refresh.This might result in metrics appearing out-of-sync and with details displayed on other pages.",
+  "metrics_lag_title": "Metrics experience lag",
   "outgoing_bytes": "Outgoing bytes ({{topic}})",
   "outgoing_bytes_all_topics": "Outgoing bytes (all topics)",
   "partition_limit": "Limit {{topic}} partitions",
--- a/locales/en/metrics.json
+++ b/locales/en/metrics.json
@@ -34,7 +34,7 @@
   "metric_kpi_topics_description": "Topics are event logs in a $t(common:kafka) instance. This metric shows the total number of topics in the $t(common:kafka) instance.",
   "metric_kpi_topics_name": "Topics",
   "metric_not_available": "Data unavailable",
-  "metrics_lag_description": "Metrics regularly experience lag, and do not automatically refresh.This might result in metrics appearing out-of-sync and with details displayed on other pages.",
+  "metrics_lag_description": "Metrics regularly experience lag, and do not automatically refresh. This might result in metrics appearing out-of-sync and with details displayed on other pages.",
   "metrics_lag_title": "Metrics experience lag",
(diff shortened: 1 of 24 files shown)


jgiardino:
Has something changed with building the storybook preview? There used to be a chromatic action listed under the checks that provided the URL for storybook previews, but I'm not able to find any URL.

locales/en/metrics.json
```diff
@@ -34,6 +34,8 @@
   "metric_kpi_topics_description": "Topics are event logs in a $t(common:kafka) instance. This metric shows the total number of topics in the $t(common:kafka) instance.",
   "metric_kpi_topics_name": "Topics",
   "metric_not_available": "Data unavailable",
+  "metrics_lag_description": "Metrics regularly experience lag, and do not automatically refresh. This might result in metrics appearing out-of-sync and with details displayed on other pages.",
```
jgiardino, line 37: One more typo here that I missed 😵‍💫 
The "and" should be removed from the second statement:
> Metrics regularly experience lag, and do not automatically refresh. This might result in metrics appearing out-of-sync ~and~ with details displayed on other pages.

riccardo-forina:
It's there, you probably only have to scroll the actions to see it as it's one of the last ones.

I know this only because I had the same problem in the past and spent a unhealthy amount of hours trying to figure out why it wasn't working before noticing GH shows only the first 6 actions... 😅

jgiardino:
These updates look great! Here are a few comments:

1. When both conditions for the selected size are true: tech preview and unavailable in the selected region
    * Currently only the "unavailable in the selected region" message displays when both conditions are true. Without any user research to back this up, my instinct is to display both messages. Knowing that the selected size is in tech preview might encourage me to select a different size. 
    * Given that we want to refactor this design, I'm also fine with putting this on hold and addressing it as part of the next design. We can also raise this in a UX review to get input from more of the team on this. 
2. improved accessibility for screen readers 
    * Any messages that display within the form for specific fields should be linked to the form inputs via `aria-describedby` and `id` attributes so that screen reader users hear these messages announced when they move focus to the form input. For example:
      * `aria-describedby="instance-size-unavailable"` on the input
      * `id="instance-size-unavailable"` on the message/alert
    * Support for `aria-describedby` on the slider was just added in https://github.com/patternfly/patternfly-react/pull/7408. Even though this version of PF isn't released yet, should we go ahead and add the `aria-describedby` [prop](https://patternfly-react-pr-7408.surge.sh/components/slider#slider) on the slider? I don't expect this to result in errors, but when we finally update PF, this association will already be set up.
    * The message that communicates the selected size is unavailable in the selected region displays below the Size input, but it also applies to the Cloud region input. Therefore the `aria-describedby` attribute should also be added to the cloud region input.

PR #377, opened 2022-05-27
chore: add a ResponsiveTable with checkbox example

riccardo-forina:

Add an example for the `ResponsiveTable` where a checkbox column is used.

### DIFF
--- a/src/shared/Table/ResponsiveTable.stories.tsx
+++ b/src/shared/Table/ResponsiveTable.stories.tsx
@@ -8,7 +8,7 @@ import {
 import { InfoIcon } from "@patternfly/react-icons";
 import { actions } from "@storybook/addon-actions";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
-import { VoidFunctionComponent } from "react";
+import { useState, VoidFunctionComponent } from "react";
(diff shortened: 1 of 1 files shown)


HemaHG:
Thanks @riccardo-forina, it's really helpfull

riccardo-forina:
Thinking more about it, don't do this. We need to natively support the checkbox in the responsive table so to handle the responsive part of the problem better.

The way it's done in the example I shared, the first column with content will be hidden if it gets small enough.

suyash-naithani:
@riccardo-forina I am using something similar in acls table. I've wrapped my rows that I wrote in components in a <TableComposable> .
A challenge with that was getting the alignment of my table headers with the respective field for different resource types. 
Do you think it would be a good idea for me to use this component for the acls?

riccardo-forina:
>Do you think it would be a good idea for me to use this component for the acls?

I'd say it's worth a try.

HemaHG:
@riccardo-forina, I have 3 table components for consumers, so I have replaced the table with the responsive table, I have one question, how to apply sort for a particular column, I tried applying sort, but it's getting applied to all the columns

suyash-naithani:
Could you rebase this PR with the latest changes?

HemaHG:
created Jira https://issues.redhat.com/browse/MGDSTRM-8793

PR #401, opened 2022-06-06
chore: Add data-ouia-component-id for create topic button

HemaHG:

> Add data-ouia-component-id for create topic button

> https://issues.redhat.com/browse/MGDSTRM-8695


nwanders:
You're right, Hema. It does specify "button-create." I was confused because I specified "button-create-topic" later in the document. After discussing with you, let's keep the simple "button-create"

riccardo-forina (approved):

PR #405, opened 2022-06-08
fix : add deprecation eslint rule

ajaypratap003:

Add deprecation eslint rule.

https://issues.redhat.com/browse/MGDSTRM-8457

**Verification steps** 
Run lint cmd : npm run lint 

  No

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,61 @@
+{
+  "root": true,
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -0,0 +1,61 @@
+{
```
riccardo-forina, line 1: There is already an `.eslintrc.js` file, please use that

riccardo-forina (changes requested):

PR #411, opened 2022-06-13
chore: kafka connection tab

HemaHG:

> Kafka connection tab Phase 1
> Kafka connection tab Phase 2

> https://docs.google.com/document/d/1GeZnF8d8k9esakNSlWC_2FUVbZWLnNFHedracLMYFTw/edit?pli=1#

**Verification steps** 
> Kafka > KafkaInstanceDrawer > KafkaConnectionTab > ConnectionTab

Phase 1
https://619e1921b3e38b003afeb2be-ihbcxldhmo.chromatic.com/?path=/story/kafka-kafkainstancedrawer-kafkaconnectiontab--connection-tab

Phase 2
https://619e1921b3e38b003afeb2be-ihbcxldhmo.chromatic.com/?path=/story/kafka-kafkainstancedrawer-kafkaconnectiontabp2--connection-tab


riccardo-forina:
Just to understand, `ConnectionTabV1` maps to variation 2 in the document?

HemaHG:
@jenn updated the design today, I am currently working on it

HemaHG:
@riccardo-forina I have made the Phase 1 changes in the existing kafkaConnectionTab file and Phase 2 changes in a new file with the name KafkaConnectionTabP2, let me know if it's a good approach or should I need to follow some other approach

HemaHG:
@jenn, I don't think using a forming group will fit this, so I have recreated the styles as you suggested, let me know your thoughts on this. I have made changes only to the `Rest API` part

jgiardino:
Your updates look great, @HemaHG! 

Can you refactor the following section to use a similar structure as the clipboard copy labels in the new REST API section? (E.g use `<strong>` instead of `<h6>` for the label)

image

wtrocki:
Screenshot 2022-06-15 at 09 47 59

Should I be seeing some urls here?

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
...
+  "kafka_service_fleet_manager_button_aria_label": "More information about the URL for Kafka Service Fleet Manager",
+  "kafka_service_fleet_manager_description": "Connect to the control plane to manage Kafka instances and service accounts, and to access metrics.",
+  "kafka_service_fleet_manager_popover_body": "Use the REST API URL to connect to the control plane to manage Kafka instances and service accounts, and to access metrics.",
+  "kafka_service_fleet_manager_popover_header": "URL for Kafka Service Fleet Manager",
+  "kafka_service_fleet_manager_url": "URL for Kafka Service Fleet Manager",
+  "more_info_about_sasl_oauthbearer": "More information about SASL/OAUTHBEARER",
+  "popover_helper_text": "For available endpoints, see the <0>API Swagger documentation</0>. For details about using this API, see the <1>API documentation</1> in the Customer Portal.",
+  "recommended": "Recommended",
+  "rest_api_description": "Streams for Apache Kafka provides two REST APIs, one for accessing the Kafka instance and one for accessing the control plane.",
```
wtrocki, line 21: Control plane is not well known term outside red hat. "Managing your kafka" is more generally understandable term

HemaHG:
@wtrocki , Rest Apis is just a header for an expandable section, clicking on it will show the URLs for Kafka admin and Kafka fleet manager

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
...
+  "bootstrap_server": "Bootstrap server",
+  "bootstrap_server_description": "Establish the initial connection between your client and Kafka instance.",
+  "create_service_account_to_generate_credentials": "Create a service account to generate credentials. Manage the service accounts you create on the <value>Service Accounts</value> page.",
+  "current_instance": "Accounts do not have access to this instance unless granted under the instance’s <value>Access tab</value>. The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.",
+  "drawer_resource_tab_body_description_1": "Your application or tool initially connects to the Kafka instance using the bootstrap server and authenticates with your service account credentials. You can also access the Kafka instance by calling the REST API using a token obtained from Red Hat SSO.",
+  "kafka_service_fleet_manager_button_aria_label": "More information about the URL for Kafka Service Fleet Manager",
+  "kafka_service_fleet_manager_description": "Connect to the control plane to manage Kafka instances and service accounts, and to access metrics.",
```
HemaHG, line 14: @wtrocki Should  also need to replace the term 'control plane' used here

wtrocki: My personal opinion: yes
Control plane is some kubernetes pattern that is not known outside kubernetes.

@jgiardino voice matters in terms of naming.

jbyrne-redhat: I agree that _control plane_ is a term we should avoid.

We can maybe solve this problem by eliminating the "Connect to..." phrase entirely:

- "Configure topics, consumer groups, and access for an individual Kafka instance."
- "Manage Kafka instances and service accounts, and access metrics."

WDYT?

jbyrne-redhat: Also, I don't see where the short description for the **REST APIs** section is defined, but we probably need to update that, too. Maybe something like the following:

"Streams for Apache Kafka provides two REST APIs, one for configuring individual Kafka instances, and one for managing resources such as Kafka instances and services accounts."

locales/en/connection-tab.json
jbyrne-redhat: +1

locales/en/connection-tab.json
HemaHG: @jbyrne-redhat, attaching the screenshot for REST APIs short description
image

jbyrne-redhat: Thanks, @HemaHG  - yeah I saw the short description in the Storybook preview that Jenn pointed me to. I saw that we're using the term _control plane_ there too, so suggested some alternative wording ^^.

HemaHG: Thanks, @jbyrne-redhat for reviewing. I have updated the PR accordingly https://619e1921b3e38b003afeb2be-csichqiapa.chromatic.com/?path=/story/kafka-kafkainstancedrawer-kafkaconnectiontab--connection-tab

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
...
+  "create_service_account_to_generate_credentials": "Create a service account to generate credentials. Manage the service accounts you create on the <value>Service Accounts</value> page.",
+  "current_instance": "Accounts do not have access to this instance unless granted under the instance’s <value>Access tab</value>. The owner, the organization administrator, or users with the Allow Alter permission on this instance can manage access.",
+  "drawer_resource_tab_body_description_1": "Your application or tool initially connects to the Kafka instance using the bootstrap server and authenticates with your service account credentials. You can also access the Kafka instance by calling the REST API using a token obtained from Red Hat SSO.",
+  "kafka_service_fleet_manager_button_aria_label": "More information about the URL for Kafka Service Fleet Manager",
+  "kafka_service_fleet_manager_description": "Manage Kafka instances and service accounts, and access metrics.",
+  "kafka_service_fleet_manager_popover_body": "Use the REST API URL to connect to manage Kafka instances and service accounts, and to access metrics.",
```
jgiardino, line 15: @wtrocki @jbyrne-redhat  - The other popover includes the following contents:
> Use the REST API URL to connect to this Kafka instance over HTTPS to manage topics, consumer groups, and access.

Should we also include the "over HTTPS" part here too? Or is this one special?

wtrocki: I think that can be skipped.
Rest implies using HTTP protocol (of which https is just special case)

wtrocki:
Root is the url users use to call service. Api file is an json file that describes api (called openapi). UI is an api UI - I put real links for managment apis do they can be clicked for more info

wtrocki:
> We are stuck with "fleet manager" for now until the name of the API is updated

We can update that very easily in kas-fleet-manager repo - what name we should use? Kafka Managment is ok? cc @dimakis

locales/en/connection-tab.json
jgiardino: Thanks, @wtrocki! Just to confirm, you're suggesting we can remove the text "over HTTPS" from the following popover?

image

wtrocki:
> For the API UI, from previous discussions, the conclusion was that within the consoledot UI, we should direct the user to where the API UI lives under https://console.redhat.com/beta/docs/api. 

Yes. Now I know what I have missed in my explanation. 
We not going to get https://console.redhat.com/beta/docs/api for that specific epic (as well as access.redhat.com documentation) there is separate epic for it. 
As long we provide api root and ability to get OpenAPI file (could be just json or api UI as above - or best both) end users will be happy. Once that is there we will have follow up to get proper api strategy for all services.

> Does the user need to be able to see the value of the OpenAPI, or just copy it?

Just copy

> Should the popover contents be displayed inline since we already have these contents in a collapsible section?

@HemaHG can help :)

wtrocki:
> For API, is that an alternate way to get the same details provided by the API UI?

Yes. Example for "Management API" https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service

wtrocki:
Thank you so much for getting all clarified. Agree to ignore API URL for the moment.

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
...
+  "kafka_service_fleet_manager_popover_header": "URL for Kafka Service Fleet Manager",
```
jgiardino, line 16: FYI, it looks like the name of this API will soon change from **Kafka Service Fleet Manager** to **Kafka Management API**
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1135

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
+{
+  "admin_url_button_aria_label": "More information about the URL for Kafka Admin REST API",
+  "admin_url_description": "Configure topics, consumer groups, and access for an individual Kafka instance.",
```
jgiardino, line 3: One minor change to what was originally suggested here. Please change "an individual" to "this", since the URL we display is specific to the selected Kafka instance.

locales/en/connection-tab.json
```diff
@@ -0,0 +1,29 @@
+{
+  "admin_url_button_aria_label": "More information about the URL for Kafka Admin REST API",
```
jgiardino, line 2: @wtrocki Following up on our chat offline, do you know what the updated name for this one will be? Is it definitely changing from **Kafka Admin REST API** to **Kafka Instance API** or is that still TBD?

wtrocki: I will create PR for Kafka Instance but that will be subjective to PR being approved.
Kafka, Connectors and Registry management were approved so now I will look into dataplane names

jgiardino:
Following a chat with @HemaHG and @riccardo-forina 

- We'll update `https://api.openshift.com/api/kafkas_mgmt/v1` to be `https://api.openshift.com/api/kafkas_mgmt/v1/openapi` in both versions
- In the second version, if the API UIs are not available in consoledot when the portal documentation becomes available, we will use an external link.
- We'll move forward with adding the first version to production.

wtrocki:
looks good

jgiardino (approved):
Thanks, @HemaHG!

PR #417, opened 2022-06-14
chore: POC add modal to warn user when they are leaving to accept T&C

Kells512:

> Warn user that they are leaving to accept T&C. When making a Kafka Instance for the first time, they are directed to new screen where they accept the T&C flow, this caused confusion where some ignored the screen and tried to continue without accepting. 

**Verification steps** 
TandCModal

>Story can be found in < POC > < SmallModal > < Story >

>There might need to be extra space between normal and bold text

### DIFF
--- a/locales/en/manage-kafka-permissions.json
+++ b/locales/en/manage-kafka-permissions.json
@@ -11,7 +11,6 @@
   "dialog_aria_label": "Manage access dialog",
   "dialog_title": "Manage access",
   "is_kafka_instance": "is '{{value}}' ",
-  "is_kafka_instane": "",
   "kafka_instance": "$t(common:kafka) instance",
   "no_results_found": "No results found",
   "operations": {
--- a/src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
+++ b/src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
@@ -1,0 +1,61 @@
+import React, { VoidFunctionComponent } from "react";
+import { Modal, ModalVariant, Button } from "@patternfly/react-core";
+
+var linebreak = "\n";
+
+export class SmallModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isModalOpen: false,
+    };
+    this.handleModalToggle = () => {
+      this.setState(({ isModalOpen }) => ({
+        isModalOpen: !isModalOpen,
+      }));
+    };
+  }
+
+  render() {
+    const { isModalOpen } = this.state;
+
+    return (
+      <React.Fragment>
+        <Button variant="primary" onClick={this.handleModalToggle}>
+          Show small modal
+        </Button>
+        <Modal
+          id="modalTerms"
+          variant={ModalVariant.small}
+          title="Terms and Conditions"
+          isOpen={isModalOpen}
+          onClose={this.handleModalToggle}
+          actions={[
+            <Button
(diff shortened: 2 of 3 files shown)


locales/en/manage-kafka-permissions.json
```diff
@@ -11,6 +11,7 @@
   "dialog_aria_label": "Manage access dialog",
   "dialog_title": "Manage access",
   "is_kafka_instance": "is '{{value}}' ",
+  "is_kafka_instane": "",
```
jgiardino, line 14: Did you mean to include this new line? It looks like it was unintentional.

src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+          ]}
+        >
+          <Flex direction={{ default: "column" }}>
+            <FlexItem>
+              <div>
+                Red Hat has Terms and Conditions for its managed services.
+              </div>
+            </FlexItem>
+            <FlexItem>
+              <div class="pf-u-font-weight-bold">
```
jgiardino, line 58: Similar as my previous comment, it would be good to use more semantic elements here. 

In this case though I think you could replace the `<div>` with `<p><strong>` elements. The `<p>` provides the semantics of this being a paragraph, but the `<strong>` provides the semantics about this being something we want emphasized. 

I don't think the `class="pf-u-font-weight-bold"` is needed if using the `<strong>` since it's redundant, but I don't think it hurts either. 🤷‍♀️

src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    return (
+      <React.Fragment>
+        <Button variant="primary" onClick={this.handleModalToggle}>
+          Show small modal
+        </Button>
+        <Modal
+          id="modalTerms"
+          variant={ModalVariant.small}
+          title="Terms and Conditions"
```
jgiardino, line 29: Any text we display belongs in a json file for translation. This text defined by the title, as well as other text strings provided below, should be moved to a json file.

As far as which json file and how they should be organized is something I'm hoping another team member can provide guidance on. 😜

locales/en/manage-kafka-permissions.json
Kells512: No I don't thinks so, I'll take that out.

src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
Kells512: I will also change it to strong

src/ProofOfConcepts/TAndCModal/tAndCModal.tsx
riccardo-forina: In the PoC is ok to have the copy like this, we'll make the translation files when we'll prepare the PoC for production.

Kells512: Who could I ask about which json file?

Kells512: Oh okay, I will go forward with the other changes.

jgiardino (approved):
Your updates look good to me, @Kells512! Thank you for making those changes.

PR #433, opened 2022-06-22
feat: Integrate all the components of Consumer Group

HemaHG:

> Integrate all the components of Consumer Group

**Verification steps** 

> Kafka > ConsumerGroups > ConsumerGroup


HemaHG:
will create a consumer group table with a new share table component

PR #435, opened 2022-06-24
fix: change epoch from string to int value

wtrocki:

When selecting my filter options I set it to Epoch timestamp. If I enter any valid number the UI displays a negative number even when using the UI buttons for iteration of the number. When typing a large number into the field (about 9 digits) the whole UI crashes and I get an error page. Note even the large number is displayed as negative when crashing.

https://issues.redhat.com/browse/MGDSTRM-8958

**Verification steps** 

Screenshot 2022-06-24 at 17 43 11

 
I wasn't sure if we actually want to start with empty string. 
I would probably put current time `new Date().getTime` as initial value instead of 
empty string. This is more UX issue so left it as it was before.

### DIFF
--- a/src/Kafka/KafkaMessageBrowser/components/FilterGroup.tsx
+++ b/src/Kafka/KafkaMessageBrowser/components/FilterGroup.tsx
@@ -147,7 +147,7 @@ export const FilterGroup: VoidFunctionComponent<FilterGroupProps> = ({
               }
             }}
             value={
-              timestamp ? formatInTimeZone(parseISO(timestamp), "UTC", "t") : ""
+              timestamp ? Math.floor(parseISO(timestamp).getTime() / 1000) : ""
             }
           />
         )}


suyash-naithani (approved):

wtrocki:
As in description - I did not fix the UX problem just code one. For UX i think we should start with current timestamp for user and then he can subtract number easily etc.

jgiardino:
Adding some comments from an off-thread chat here for easy reference:

The search button that displays is meant to be the form submit button for all the selection criteria that display in the toolbar. Currently, if I change the values in the toolbar, the search button becomes enabled so that I can apply my updates to the view. So that button should be an explicit way for the user to say "I'm done, use this" but I also like the idea of using onBlur to trigger the display of the invalid format message. 

I also think Enter key for any of these inputs should be tied to that search button.

image

But the search button is not very prominent. Especially in smaller viewports where the filter icon toggles the display of these, the search stays in the toolbar, but maybe it should also be moved to the section that displays with the toggle.

Some enhancements we discussed:
* make that Search 🔍 button a primary button
* In smaller viewports, stack the button underneath the selection criteria, and display the label "Search" along with the 🔍  icon.

riccardo-forina:
Thank you for this @wtrocki, as discussed offline we'll try to tackle this error from a UX PoV, changing a bit the state machine that powers the system.

PR #445, opened 2022-07-04
chore: add peer dependency xstate/inspect

anukritijha:

> Add missing peer dependency for xstate/inspect


package.json
```diff
@@ -111,6 +111,7 @@
   },
   "dependencies": {
     "@swc/helpers": "0.4.3",
+    "@xstate/inspect": "^0.7.0",
```
riccardo-forina, line 114: It should be a dev deps

PR #451, opened 2022-07-06
[WIP] chore: Display the consumer group drawer section content using description list

hemahg:

> Update the consumer group drawer section to use the description list to display the content

> fixes https://issues.redhat.com/browse/MGDSTRM-7703


HemaHG:
This is not the expected implementation

PR #455, opened 2022-07-08
chore: Delete Kafka Instance, Topic and Consumer Group

Kells512:

> make updates to Delete modals based on usability findings

**Verification steps** 

>Delete Kafka Instance:

Screenshot from 2022-07-01 10-22-03

>  Kafka>  < DeleteKafkaInstance > < Story >

>Delete Kafka Topic (Without Active users):
Screenshot from 2022-07-01 10-24-09

>  Kafka>  < DeleteKafkaTopic > < Story >

>Delete Kafka Topic (With Active users):

Screenshot from 2022-07-01 10-25-24

>  Kafka>  < DeleteKafkaTopicActiveUser > < DeleteKafkaTopicActive > < Story >

>Delete Consumer Group
Screenshot from 2022-07-01 11-01-19

>  Kafka>  < ConsumerGroups >  < components > < DeleteConsumerGroup > < Allow Consumer Group Deletion >

**Don't have service account done yet because it doesn't have a place to go yet**


locales/en/kafka.json
```diff
@@ -82,6 +88,11 @@
     "uncosumed_partitions_for_topic": "Unconsumed partitions for this topic",
     "view_partitions_offsets": "View partition offsets"
   },
+  "delete_instance_message": "The <0>{{instanceName}}</0> Kafka instance and all topics will be deleted. Applications will no longer have access to the topics in this instance.",
+  "delete_instance_title": "Delete instance?",
+  "delete_topic_active_message": "One or more consumers are currently active in this topic, and will be removed at the time of deletion. ",
+  "delete_topic_message": "The <0>{{instanceName}}</0> topic will be deleted and removed from this instance. Applications will no longer have access to this topic.",
+  "delete_topic_title": "Delete topic",
```
jgiardino, line 95: I think this string should have a "?" at the end. There is one missing in the modal title:
image

jgiardino:
Thanks for making these updates. I posted a comment inline regarding some text string changes.

While reviewing these, I realized that the shared Delete modal you’re basing the new components on could use some refactoring. I’m not sure if that’s something that should be addressed as part of this PR, or not. 

In the original implementation of the shared Delete modal, there are two instances of the `pf-c-modal-box__body` section that are nested. This results is more padding around the contents than we should have. I also don't see any examples in PF showing that you could have multiple body elements as siblings in the modal. I think all contents in the modal are expected to reside in the same body. 

Making this change will also require some additional layout within the body of Kafka delete modals to add the space we need between contents. I had created a [codesandbox](https://codesandbox.io/s/modaldelete-lpwi0?file=/index.js) a while back to explore how these contents could be implemented. There are probably some things in there that we don't want to reuse, but it might be helpful to see what some options are. I _think_ the right thing to do is to use a [stack layout with a gutter](https://www.patternfly.org/v4/layouts/stack#with-gutter) to create space between the message text and form input with label.

@riccardo-forina - What would you recommend here?

locales/en/kafka.json
Kells512: Yeah I will fix that now

riccardo-forina:
I'd say that I originally disabled the default body wrapper in the shared delete, and then enabled it back but I can't recall exactly why. Probably it was by mistake. I'd suggest flipping [this boolean](https://github.com/redhat-developer/app-services-ui-components/blob/main/src/shared/DeleteModal/DeleteModal.tsx#L164) and see what happens, it should make it work properly.

riccardo-forina:
@jgiardino you should now see the link for storybook

jgiardino:
Thanks, @riccardo-forina. I checked again, but I still see the issue with the spacing. But I'm also not sure if @Kells512 tried your suggestion or not with her latest commits.

riccardo-forina:
Ah, no, that's still on the pen I think.

Kells512:
> Thanks, @riccardo-forina. I checked again, but I still see the issue with the spacing. But I'm also not sure if @Kells512 tried your suggestion or not with her latest commits.

Yes I will do that today

Kells512:
> Thanks, @riccardo-forina. I checked again, but I still see the issue with the spacing. But I'm also not sure if @Kells512 tried your suggestion or not with her latest commits.

It is done now :)

jgiardino:
Thank you @Kells512! Your examples look good to me.

riccardo-forina (approved):

riccardo-forina:
@Kells512 it looks like by flipping that boolean we need to manually set an aria attribute

```
Modal: When using hasNoBodyWrapper or setting a custom header, ensure you assign an accessible name to the the modal container with aria-label or aria-labelledby.
```

Ping me in chat if you need help figuring this out

PR #459, opened 2022-07-11
chore: Remove service account translation

HemaHG:

> Moving the strings in the service account translation file to the Kafka file, since the strings are related to KafkaConnectionTab.

### DIFF
--- a/src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
+++ b/src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
@@ -71,7 +71,7 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
       </TextContent>
       <TextContent className="pf-u-pb-sm">
         <Text component={TextVariants.h3} className="pf-u-mt-xl">
-          {t("service-account:service_accounts_small")}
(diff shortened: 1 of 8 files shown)


src/Kafka/KafkaInstanceDrawer/KafkaConnectionTab.tsx
```diff
@@ -71,7 +71,7 @@ export const KafkaConnectionTab: FunctionComponent<KafkaConnectionTabProps> = ({
       </TextContent>
       <TextContent className="pf-u-pb-sm">
         <Text component={TextVariants.h3} className="pf-u-mt-xl">
-          {t("service-account:service_accounts_small")}
+          {t("kafka:connection_tab.service_accounts_small")}
```
riccardo-forina, line 74: Can we use the connection-tab namespace here too? It would keep all these strings in a single place. But there might be a good reason you did this and I don't realize it

HemaHG: The translation for the connection tab was already there in the kafka.json file. Since there were two versions of the Kafka connection tab that needed to be created, I created the connection-tab.json file and Used kafka.json for v1, the connection-tab.json file for v2

riccardo-forina (approved):

PR #461, opened 2022-07-12
chore: adding AWS link and Add GCP to the Kafka overview page

Kells512:

> Replace this with a quick explanation of what's included here. 

**Verification steps** 


src/AppServices/KafkaPageV3.tsx
```diff
@@ -0,0 +1,307 @@
...
+              </Flex>
+            </Flex>
+          </CardBody>
+        </Card>
+      </MarketingPageSection>
+    </>
+  );
+};
+
+const SupportLink: FunctionComponent = ({ children }) => (
```
HemaHG, line 281: There is a shared component for creating external links https://redhat-developer.github.io/app-services-ui-components/?path=/story/shared-externallink--external-link. I think we can make use of that instead of creating new components for each external link

PR #469, opened 2022-07-18
feat: Settings tab

HemaHG:

> Create UI for the settings tab

> fixes https://issues.redhat.com/browse/MGDSTRM-8810

**Where to find story**

Interactive story
https://619e1921b3e38b003afeb2be-giggariile.chromatic.com/?path=/story/kafka-settings--interactive-example&globals=withInsightsChrome:true

### DIFF
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -130,9 +130,8 @@
       "turnoff": "Off",
       "turnon": "On"
     },
-    "connection_re_authentication_title": "Connection re-authentication:",
     "error_alert_title": "Something went wrong",
-    "error_alert_title_description": "We're unable to update connection re-authentication at this time, Try again later.",
+    "error_alert_title_description": "We're unable to update connection re-authentication at this time. Try again later.",
     "kafka_instance_settings_label": "Configure global settings for this Kafka instance.",
     "success_alert": "Connection re-authentication turned {{status}}",
     "turn_off_button_label": "Turn off",
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -127,7 +127,7 @@
     "Connection_re_authentication_states": {
       "turning_off": "Turning off...",
       "turning_on": "Turning on...",
-      "turnoff": "off",
+      "turnoff": "Off",
       "turnon": "On"
     },
     "connection_re_authentication_switch": "Connection re-authentication.",
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -123,14 +123,13 @@
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
   "settings": {
-    "connection_re_authentication_label": "Connection re-authentication.",
+    "connection_re_authentication_label": "Connection re-authentication",
     "Connection_re_authentication_states": {
       "turning_off": "Turning off...",
       "turning_on": "Turning on...",
       "turnoff": "Off",
       "turnon": "On"
     },
-    "connection_re_authentication_switch": "Connection re-authentication.",
     "connection_re_authentication_title": "Connection re-authentication:",
     "error_alert_title": "Something went wrong",
     "error_alert_title_description": "We're unable to update connection re-authentication at this time, Try again later.",
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -123,11 +123,22 @@
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
   "settings": {
-    "connection_re_authentication_label": "Connection re-authentication:",
-    "disable": "Disable",
-    "kafka_instance_settings_label": "Configure global settings for the kafka instance",
-    "warning_description": "There are some potential risks",
-    "warning_title": "Disable re-authentication?"
+    "connection_re_authentication_label": "Connection re-authentication.",
+    "Connection_re_authentication_states": {
+      "turning_off": "Turning off...",
+      "turning_on": "Turning on...",
+      "turnoff": "off",
+      "turnon": "On"
+    },
+    "connection_re_authentication_switch": "Connection re-authentication.",
+    "connection_re_authentication_title": "Connection re-authentication:",
+    "error_alert_title": "Something went wrong",
+    "error_alert_title_description": "We're unable to update connection re-authentication at this time, Try again later.",
+    "kafka_instance_settings_label": "Configure global settings for this Kafka instance.",
+    "success_alert": "Connection re-authentication turned {{status}}",
+    "turn_off_button_label": "Turn off",
+    "warning_description": "There are potential risks associated with turning off re-authentication, where attackers could stay connected indefinitely.",
+    "warning_title": "Turn off re-authentication?"
   },
   "short_description": "Short Description",
   "storage": "Storage",
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -125,12 +125,12 @@
   "settings": {
     "connection_re_authentication_label": "Connection re-authentication",
     "Connection_re_authentication_states": {
+      "off": "Off",
+      "on": "On",
       "turning_off": "Turning off...",
-      "turning_on": "Turning on...",
-      "turnoff": "Off",
-      "turnon": "On"
+      "turning_on": "Turning on..."
     },
-    "error_alert_title": "Something went wrong",
+    "error_alert_title": "$t(common:something_went_wrong)",
     "error_alert_title_description": "We're unable to update connection re-authentication at this time. Try again later.",
     "kafka_instance_settings_label": "Configure global settings for this Kafka instance.",
     "success_alert": "Connection re-authentication turned {{status}}",
--- a/locales/en/kafka.json
+++ b/locales/en/kafka.json
@@ -122,6 +122,13 @@
   "partitions": "Partitions",
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication:",
+    "disable": "Disable",
+    "kafka_instance_settings_label": "Configure global settings for the kafka instance",
+    "warning_description": "There are some potential risks",
+    "warning_title": "Disable re-authentication?"
+  },
   "short_description": "Short Description",
   "storage": "Storage",
   "storage_value": "up to 1000 GiB",
--- a/src/Kafka/Settings/Settings.stories.tsx
+++ b/src/Kafka/Settings/Settings.stories.tsx
@@ -35,7 +35,7 @@ export const InteractiveExample: ComponentStory<typeof Settings> = (args) => {
       setIsModalOpen(true);
     } else {
       setConnectionStatus("TurningOn");
-      apiError<{ ConfigurationOff: boolean }>(
+      fakeApi<{ ConfigurationOff: boolean }>(
         {
           ConfigurationOff: true,
         },
--- a/src/Kafka/Settings/Settings.stories.tsx
+++ b/src/Kafka/Settings/Settings.stories.tsx
@@ -32,7 +32,7 @@ const Template: ComponentStory<typeof Settings> = () => {
       setConnectionStatus("TurningOn");
       apiError<{ ConfigurationOff: boolean }>(
         {
-          ConfigurationOff: undefined,
+          ConfigurationOff: true,
         },
         4000
       )
--- a/src/Kafka/Settings/Settings.stories.tsx
+++ b/src/Kafka/Settings/Settings.stories.tsx
@@ -35,9 +35,9 @@ export const InteractiveExample: ComponentStory<typeof Settings> = (args) => {
       setIsModalOpen(true);
     } else {
       setConnectionStatus("TurningOn");
-      fakeApi<{ ConfigurationOff: boolean }>(
+      fakeApi<{ isConnected: boolean }>(
         {
-          ConfigurationOff: true,
+          isConnected: true,
         },
         4000
       )
@@ -56,9 +56,9 @@ export const InteractiveExample: ComponentStory<typeof Settings> = (args) => {
   const onClickTurnOff = () => {
     setIsModalOpen(false);
     setConnectionStatus("TurningOff");
-    fakeApi<{ ConfigurationOff: boolean }>(
+    fakeApi<{ isConnected: boolean }>(
       {
-        ConfigurationOff: false,
+        isConnected: false,
       },
       4000
     )
@@ -122,8 +122,8 @@ TurnONConnectionReauthentication.args = {
   connectionState: true,
 };
 
-export const TurnONConnectionReauthenticationFail = Template.bind({});
-TurnONConnectionReauthenticationFail.args = {
+export const TurnOnConnectionFailure = Template.bind({});
+TurnOnConnectionFailure.args = {
   connectionStatus: "Off",
   alertStatus: "danger",
 };
--- a/src/Kafka/Settings/Settings.stories.tsx
+++ b/src/Kafka/Settings/Settings.stories.tsx
@@ -3,6 +3,7 @@ import { AlertVariant } from "@rhoas/app-services-ui-shared";
 import type { ComponentStory, ComponentMeta } from "@storybook/react";
 import { useState } from "react";
 import { Settings } from "./Settings";
+import { fakeApi } from "../../shared/storiesHelpers";
 
 export default {
   component: Settings,
@@ -10,12 +11,13 @@ export default {
 } as ComponentMeta<typeof Settings>;
 
 const Template: ComponentStory<typeof Settings> = () => {
-  const isSettingsChangeable = "success";
   const [isChecked, setIschecked] = useState<boolean>(true);
 
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
 
   const [showAlert, setShowAlert] = useState<AlertProps[]>([]);
+
+  const [isLoading, setIsLoading] = useState<boolean>(false);
 
   const onChange = () => {
     setIsModalOpen(true);
@@ -31,21 +33,32 @@ const Template: ComponentStory<typeof Settings> = () => {
   };
 
   const onDisable = () => {
-    setIschecked(false);
     setIsModalOpen(false);
-    isSettingsChangeable === "success"
-      ? addAlert(
+    setIschecked(false);
+    setIsLoading(true);
+    fakeApi<{ ConfigurationOff: boolean }>(
+      {
+        ConfigurationOff: false,
+      },
+      2000
+    )
+      .then(() => {
+        addAlert(
           "Connection re-authentication is turrning off",
           AlertVariant.success,
           "Broker is restarting to apply the new configurations. This process might take several minutes",
           "success-alert"
-        )
-      : addAlert(
+        ),
+          setIsLoading(false);
+      })
+      .catch(() =>
+        addAlert(
           "Something went wrong",
           AlertVariant.danger,
           "We're unable to update connection re-authentication at this time, Try again later.",
           "danger-alert"
-        );
+        )
+      );
   };
 
   const onClose = () => {
@@ -65,7 +78,7 @@ const Template: ComponentStory<typeof Settings> = () => {
       onClose={onClose}
       showAlert={showAlert}
       closeAlertAction={closeAlertAction}
-      isSettingsChangeable={isSettingsChangeable}
+      isLoading={isLoading}
     />
   );
 };
--- a/src/Kafka/Settings/Settings.stories.tsx
+++ b/src/Kafka/Settings/Settings.stories.tsx
@@ -2,10 +2,13 @@ import type { ComponentStory, ComponentMeta } from "@storybook/react";
 import { useState } from "react";
 import { Settings } from "./Settings";
 import { apiError, fakeApi } from "../../shared/storiesHelpers";
-import type { SettingsStatus } from "./types";
+import type { AlertStatus, SettingsStatus } from "./types";
 
 export default {
   component: Settings,
+  args: {
+    connectionStatus: "On",
+  },
   parameters: {
     backgrounds: {
       default: "Background color 100",
@@ -13,15 +16,17 @@ export default {
   },
 } as ComponentMeta<typeof Settings>;
 
-const Template: ComponentStory<typeof Settings> = () => {
+const Template: ComponentStory<typeof Settings> = (args) => (
+  <Settings {...args} />
+);
+
+export const InteractiveExample: ComponentStory<typeof Settings> = (args) => {
   const [connectionStatus, setConnectionStatus] =
     useState<SettingsStatus>("On");
 
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
 
-  const [alertStatus, setAlertStatus] = useState<
-    "success" | "failure" | undefined
-  >(undefined);
+  const [alertStatus, setAlertStatus] = useState<AlertStatus>();
 
   const [connectionState, setConnectionState] = useState<boolean>(false);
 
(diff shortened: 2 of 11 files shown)


locales/en/kafka.json
```diff
@@ -109,6 +109,13 @@
   "partitions": "Partitions",
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication:",
+    "disable": "Disable",
```
jgiardino, line 114: Assuming this is the label for the submit button in the prompt, can you change this to "Turn off"?

Chatting with the CCS team, they recommend against using "Disable" for the prompt:

- IBM Style Guide also recommends against disable in this case: 
https://www.ibm.com/docs/en/ibm-style?topic=word-usage#d
- Everywhere else, we use the language "Turn off", "Turn on" or some variation of this

locales/en/kafka.json
```diff
@@ -109,6 +109,13 @@
...
+    "kafka_instance_settings_label": "Configure global settings for the kafka instance",
+    "warning_description": "There are some potential risks",
+    "warning_title": "Disable re-authentication?"
```
jgiardino, line 117: Similar to previous comment, please change this text to:
> Turn off re-authentication?

locales/en/kafka.json
```diff
@@ -109,6 +109,13 @@
   "partitions": "Partitions",
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication:",
+    "disable": "Disable",
+    "kafka_instance_settings_label": "Configure global settings for the kafka instance",
+    "warning_description": "There are some potential risks",
```
jgiardino, line 116: After chatting with the CCS team, here is an updated message to use:
> There are potential risks associated with turning off re-authentication, where bad actors could stay connected indefinitely.

Please note:
- There is an outstanding question in the [documentation PR](https://github.com/redhat-developer/app-services-guides/pull/553/files#r933265729) regarding whether we should use something other than "bad actor". Any decision made in that PR should ultimately be applied to this message, too. This could be a future PR, but also possible that this will be resolved while this PR is open.
- FYI, in the future, when the documentation is published, we will want to add a link "Learn more" to access the documentation. This will be covered as part of a follow-up task: https://issues.redhat.com/browse/MGDSTRM-8809

src/Kafka/Settings/Settings.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+      {
+        ConfigurationOff: false,
+      },
+      2000
+    )
+      .then(() => {
+        addAlert(
+          "Connection re-authentication is turrning off",
+          AlertVariant.success,
+          "Broker is restarting to apply the new configurations. This process might take several minutes",
```
jgiardino, line 49: A few notes about this:
1. These strings should be moved to the translation file. 
2. A success should display when changing the value of this setting to either `off` or `on`, and not just when changing the value to `off`.
3. The message should only display when the API response to changing the value is successful
4. The message should not provide important details about how long it takes for settings to fully take affect.

Following the points above, there are some text string changes needed in the toast alert:

**Re: points 2. & 3.**, the alert title should change to one of the following depending on the value change:

- If turning off the setting:
    > Connection re-authentication turned off
- If turning on the setting: 
    > Connection re-authentication turned on

**Re: point 4.**, the message body text should be removed (i.e. "Broker is restarting...")

locales/en/kafka.json
```diff
@@ -109,6 +109,13 @@
   "partitions": "Partitions",
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication:",
```
jgiardino, line 113: Should the ":" be included as part of this text string? I'm not sure if there's any benefit to the translator. We also should reuse this label for an aria-label where I wouldn't expect to include punctuation like this.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        "Off"
+                      )}
+                    </FlexItem>
+                  </Flex>
+                </LevelItem>
+                <LevelItem>
+                  <Switch
+                    id="Connection-re-authentication-switch"
+                    isChecked={isChecked}
+                    onChange={onChange}
```
jgiardino, line 89: Please add `aria-label` here. It should have the text string `connection_re_authentication_label` as the value.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+        </Card>
+      </PageSection>
+      <Page>
+        <PageSection isWidthLimited isCenterAligned>
+          <Card>
+            <CardBody>
+              <Level>
+                <LevelItem>
+                  <Flex>
+                    <FlexItem spacer={{ default: "spacerNone" }}>
```
jgiardino, line 62: Using `default: "spacerNone"` here results in no space between the label and the value, which looks odd:
image

Use the `sm` spacer instead. That will add some space as shown below:
image

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      <Title headingLevel={"h3"}>
```
jgiardino, line 63: Ideally, the text that represents the setting label and value would be inline elements grouped together within the same block-level element so that the text is treated as one block of text (i.e. read aloud as one block of text by screen readers). This helps screen reader users easily know what the current value of the setting is when they're reading this text.

For example:
`<strong>Connection re-authentication:</strong> On`

The `<strong>` element here provides emphasis on the label. This provides both visual separation between the label and value for sighted users, but also this emphasis helps differentiate these strings when announced by a screen reader. 

To provide this text as a single block, you can make these changes:
- On the `<FlexItem>` add `component="span"` to make these inline elements instead of block elements.
- Replace the `<Title>` component with a `<strong>` element. The Title component results in a block level element, whereas `<strong>` is an inline element.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    </FlexItem>
+                    <FlexItem>
+                      {isChecked ? (
+                        "On"
+                      ) : isLoading ? (
+                        <>
+                          <Spinner
+                            size="md"
+                            aria-valuetext={"Turning off..."}
+                            aria-label={"Turning off..."}
```
jgiardino, line 75: A spinner is a component that provides status. The `aria-valuetext` is the attribute that communicates what the status is, but the `aria-label` should not. Instead, `aria-label` should describe what we're providing status for, which is the setting. 

Therefore, the value for `aria-label` here should reuse the label of the setting, e.g. `connection_re_authentication_label`

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      </Title>
+                    </FlexItem>
+                    <FlexItem>
+                      {isChecked ? (
+                        "On"
+                      ) : isLoading ? (
+                        <>
+                          <Spinner
+                            size="md"
+                            aria-valuetext={"Turning off..."}
```
jgiardino, line 74: This is user-facing text that should be moved to the translation file.

We also want this text to reflect what value we are switching to, `on` or `off`, so there should be 2 separate strings depending on the desired state:
- "Turning on..."
- "Turning off..."

jgiardino:
The UI looks great. There were a few updates I posted regarding final messaging, or other text string changes. There were also a few things I noted regarding accessibility that would be good to address.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,148 @@
...
+      <PageSection>
+        <Card>
+          <CardBody>{t("settings.kafka_instance_settings_label")}</CardBody>
+        </Card>
+      </PageSection>
+      <Page>
+        <PageSection isWidthLimited isCenterAligned>
+          <Card>
+            <CardBody>
+              <Level>
```
jgiardino, line 59: The `Level` layout doesn't provide the modifiers needed to implement the layout we want for smaller viewports [shown here](https://docs.google.com/document/d/1hO_aMtRV75F48d4PV1j_LY1GMVXgFKwANeUwvJTzMoQ/edit?disco=AAAAavQirGA). In your implementation, the switch wraps underneath the text, but in that example, the switch is always on the right and only the text wraps.

I checked with one of the PF UI devs for suggestions, and this was his recommendation:
> I would just use a flex layout set to nowrap with the text flex child set to flex grow - or you could use a split layout with the text flex child set to "fill" (.pf-m-fill or the isFilled react prop)

HemaHG:
Thanks for reviewing the PR @jgiardino, I will make the updates as suggested

locales/en/kafka.json
jgiardino: FYI, the documentation PR mentioned above now uses "attacker" instead of "bad actor". Following their updates, we would use this for our message instead of the version in my previous comment:
> There are potential risks associated with turning off re-authentication, where attackers could stay connected indefinitely.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <Card>
+          <CardBody>{t("settings.kafka_instance_settings_label")}</CardBody>
+        </Card>
+      </PageSection>
+      <Page>
+        <PageSection>
+          <Card>
+            <CardBody>
+              <Flex>
+                <FlexItem component={"span"} spacer={{ default: "spacerSm" }}>
```
jgiardino, line 52: I previously shared this feedback from PF dev on how to handle the responsive layout. 

> I would just use a flex layout set to nowrap with the text flex child set to flex grow - or you could use a split layout with the text flex child set to "fill" (.pf-m-fill or the isFilled react prop)

These were the suggestions they made which I don't see applied here:
- use the nowrap prop on the flex layout
- use the flex grow prop on the flex item that contains the text
  - Also note, all text should be in a single flex item so that the text contents wrap as shown  in [the responsive example](https://docs.google.com/document/d/1hO_aMtRV75F48d4PV1j_LY1GMVXgFKwANeUwvJTzMoQ/edit?disco=AAAAavQirGA). Instead, you have separate text strings in separate flex items. 

The exact props to use should be noted in the PF react component documentation for the flex layout. If you're struggling to figure out what props to use where, please reach out.

locales/en/kafka.json
```diff
@@ -122,6 +122,24 @@
...
+      "turning_on": "Turning on...",
+      "turnoff": "Off",
+      "turnon": "On"
+    },
+    "connection_re_authentication_switch": "Connection re-authentication.",
+    "connection_re_authentication_title": "Connection re-authentication:",
+    "error_alert_title": "Something went wrong",
+    "error_alert_title_description": "We're unable to update connection re-authentication at this time, Try again later.",
+    "kafka_instance_settings_label": "Configure global settings for this Kafka instance.",
+    "success_alert": "Connection re-authentication turned {{status}}",
```
jgiardino, line 138: When this message displays, "Off" was displayed as the status. The `{{status}}` string that displays here should be lowercase instead. 

Also, I did not see a toast alert display when turning this on, but the setting also didn't update to show the on state. I'm not sure if that's an issue or just how the storybook is behaving.

locales/en/kafka.json
```diff
@@ -122,6 +122,24 @@
...
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication.",
+    "Connection_re_authentication_states": {
+      "turning_off": "Turning off...",
+      "turning_on": "Turning on...",
+      "turnoff": "Off",
+      "turnon": "On"
+    },
+    "connection_re_authentication_switch": "Connection re-authentication.",
+    "connection_re_authentication_title": "Connection re-authentication:",
```
jgiardino, line 134: You have three instances of "Connection re-authentication" defined here. Are all of these necessary? Two of these are being used for aria-labels. In these cases, the aria-label should be a match for the visible text, which means we should be able to use a single string going forward. 

Something to note is that this pattern should eventually become a shared component that is reused, either here for more settings or for other services. Having a simpler way to specify a single string that's used in all the places would help avoid issues with using this pattern in the future.

cc'ing @riccardo-forina in case I'm saying anything that is way off here.

jgiardino: Also, the first two instances of this text string include a period (.) at the end but typically we don't include periods in labels, only for complete sentences.

The other question I have is whether we should include the colon (:) in the third instance. That's not necessary for translation, and I wonder if it would be better to move that out of the text file and bake it into the UI pattern instead.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+                </FlexItem>
+                <FlexItem component={"span"} align={{ default: "alignRight" }}>
+                  <Switch
+                    id="Connection-re-authentication-switch"
+                    aria-label={t(
+                      "settings.connection_re_authentication_switch"
+                    )}
+                    isChecked={
+                      connectionStatus === "On" ||
+                      connectionStatus === "TurningOff"
```
jgiardino, line 115: Does this mean that if the status is "TurningOff" or "TurningOn" that I can click the switch again? Should we set a disabled prop if connection status is "TurningOff" or "TurningOn" while we wait for the API response?

src/Kafka/Settings/Settings.tsx
jgiardino: One follow-up comment on this, you can have a flex layout nested inside a flex layout. For example:
```
<Flex> 👈 use the nowrap prop here
  <FlexItem> 👈 use the flex grow prop here
    <Flex>
      <FlexItem component={"span"} spacer={{ default: "spacerSm" }}>...</FlexItem> 
       {include all text flex items here}
    </Flex>
  </FlexItem>
  <FlexItem> {include switch here} </FlexItem>
</Flex>
```

jgiardino:
Thanks for making the updates, @HemaHG! This latest update looks really good!! I posted a couple of comments. Some of my comments would be good to get feedback from other UI devs on just to make sure I'm not sending you in the wrong direction.

locales/en/kafka.json
HemaHG: @jgiardino looking into the doc https://docs.google.com/presentation/d/1cq6TnLYjntOZZYmXoAgCO9YuTlPRM8-VzqxbBgRi5EI/edit#slide=id.g132fbec33c0_0_23 `Connection re-authentication:` seems like a single string so I have included colon(:) in the translation file

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
...
+                  <Switch
+                    id="Connection-re-authentication-switch"
+                    aria-label={t(
+                      "settings.connection_re_authentication_label"
+                    )}
+                    isChecked={connectionStatus === "On"}
+                    isDisabled={
+                      connectionStatus === "TurningOff" ||
+                      connectionStatus === "TurningOn"
+                    }
```
jgiardino, line 124: I'm seeing inconsistent behavior here:

* When I toggle the switch to the **off** state, I see a disabled switch in the **off** state while the spinner displays, then I see an enabled switch in the **off** state.
* When I toggle the switch to the **on** state, I see a disabled switch in the **off** state while the spinner displays, then I see an enabled switch in the **on** state.

The disabled state is always displaying in the **off** state for both cases.

Instead, it would be good to display the disabled state showing the previous value while the spinner displays, then update it to the new value.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
...
+  return (
+    <>
+      <PageSection>
+        <Card>
+          <CardBody>{t("settings.kafka_instance_settings_label")}</CardBody>
+        </Card>
+      </PageSection>
+      <Page>
+        <PageSection>
+          <Card>
```
jgiardino, line 49: This could be a follow-up PR, but the original designs for this page set a max-width for the card. 

The only support PF provides for setting a max-width is on the Form component, which you can see in [this example](https://www.patternfly.org/v4/components/form#limit-width). I think there are two options for applying this style, and I'm not sure which would be the preferred method:

1. Wrap the card in a Form component with the limit width prop. 
2. Create custom CSS that applies a similar style to the card: `max-width: 55rem`

locales/en/kafka.json
```diff
@@ -122,6 +122,23 @@
...
+    "connection_re_authentication_label": "Connection re-authentication",
+    "Connection_re_authentication_states": {
+      "turning_off": "Turning off...",
+      "turning_on": "Turning on...",
+      "turnoff": "Off",
+      "turnon": "On"
+    },
+    "connection_re_authentication_title": "Connection re-authentication:",
+    "error_alert_title": "Something went wrong",
+    "error_alert_title_description": "We're unable to update connection re-authentication at this time, Try again later.",
```
jgiardino, line 135: The comma (,) after "time" should be a period (.) instead.

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { t } = useTranslation("kafka");
+
+  return (
+    <>
+      <PageSection>
+        <Card>
+          <CardBody>{t("settings.kafka_instance_settings_label")}</CardBody>
+        </Card>
+      </PageSection>
+      <Page>
```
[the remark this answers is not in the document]
jgiardino: Good catch, @suyash-naithani! I think there should only be one `<Page>` component on the page, and `<PageSection>` should only exist as a direct child in the `<Page>` component. I don't remember if `<Page>` is already provided in the product UI of not.

locales/en/kafka.json
jgiardino: I think this question comes down to how do we abstract this UI into a reusable pattern and also make this UI easy to maintain. 

Displaying the colon (:) is part of the pattern, and also isn't something that needs to be translated. To reduce the risk of inconsistency, ideally we would separate what the dev needs to specify when reusing this pattern, and not require that they provide the colon as part of the string. 

I would also reduce the number of instances in the translation file that include the same exact string. This requires more effort if changes are needed, but also increases the risk that changes will be made on one string, but not the other.

locales/en/kafka.json
```diff
@@ -122,6 +122,22 @@
   "partitions": "Partitions",
   "partitions_value": "up to 500",
   "select_filter_category": "Select filter category",
+  "settings": {
+    "connection_re_authentication_label": "Connection re-authentication",
+    "Connection_re_authentication_states": {
+      "turning_off": "Turning off...",
+      "turning_on": "Turning on...",
+      "turnoff": "Off",
```
ajaypratap003, line 130:
```suggestion
      "off": "Off",
```

locales/en/kafka.json
```diff
@@ -122,6 +122,22 @@
...
+      "turnon": "On"
```
ajaypratap003, line 131:
```suggestion
      "on": "On"
```

locales/en/kafka.json
```diff
@@ -122,6 +122,22 @@
...
+    },
+    "error_alert_title": "Something went wrong",
```
ajaypratap003, line 133:
```suggestion
    "error_alert_title": "$t(common:something_went_wrong)",
```

src/Kafka/Settings/Settings.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    expect(
+      comp.getByRole("checkbox", { name: "Connection re-authentication" })
+    ).toBeChecked();
+    expect(
+      await comp.findByText("Connection re-authentication turned on")
+    ).toBeInTheDocument();
+
+    userEvent.click(
+      comp.getByRole("button", {
+        name: "Close Success alert: alert: [object Object]",
```
ajaypratap003, line 108: It should not be `[object Object]`

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  return (
+    <>
+      <PageSection>
+        <Card>
+          <CardBody>{t("settings.kafka_instance_settings_label")}</CardBody>
+        </Card>
+      </PageSection>
+      <PageSection>
+        <Card className={"settings-card--max-width"}>
```
ajaypratap003, line 48:
```suggestion
        <Card className={"mas--settings__card"}>
```

src/Kafka/Settings/Settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  <FlexItem component={"span"} spacer={{ default: "spacerSm" }}>
+                    <strong>
+                      {t("settings.connection_re_authentication_label")}
+                      {":"}
+                    </strong>
+                  </FlexItem>
+                  <FlexItem>
+                    {(() => {
+                      switch (connectionStatus) {
+                        case "On":
```
ajaypratap003, line 62: It would be better to create enum type for connection  statuses i.e. On. Off, Turning etc.

src/Kafka/Settings/components/SettingsAlert.tsx
```diff
@@ -0,0 +1,76 @@
...
+                      values={{
+                        status: connectionState ? "on" : "off",
+                      }}
+                    />
+                  }
+                  actionClose={<AlertActionCloseButton onClose={closeAction} />}
+                  timeout={8000}
+                />
+              );
+            } else {
```
ajaypratap003, line 52: why do we need this else case ? if not then remove it.

src/Kafka/Settings/components/SettingsAlert.tsx
```diff
@@ -0,0 +1,76 @@
...
+                <Alert
+                  role={"alert"}
+                  variant={AlertVariant.danger}
+                  title={t("settings.error_alert_title")}
+                  actionClose={<AlertActionCloseButton onClose={closeAction} />}
+                >
+                  {t("settings.error_alert_title_description")}
+                </Alert>
+              );
+            } else {
```
ajaypratap003, line 67: why do we need this else case ? if not then remove it.

src/Kafka/Settings/components/SettingsAlert.tsx
```diff
@@ -0,0 +1,76 @@
...
+    useState<boolean>(false);
+
+  const closeAction = () => {
+    setIsCloseActionClicked(true);
+  };
+
+  return (
+    <AlertGroup isToast>
+      {(() => {
+        switch (alertStatus) {
```
ajaypratap003, line 33: As we have limited two cases so we can use if else case instead.
```
if(alertStatus==="alertStatus" && !isCloseActionClicked){
  //code here
}else if(alertStatus==="danger" && !isCloseActionClicked){
// code here
}
```

ajaypratap003 (changes requested):
Looks good. Added few inline comments/suggestions.

src/Kafka/Settings/Settings.tsx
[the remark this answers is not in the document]
ajaypratap003: > This componenent might cause nesting of page when this is integrated. @ajaypratap003 wdyt?

Yes, we should not have Page here if we have it in parent component.

src/Kafka/Settings/Settings.tsx
HemaHG: I have created the type to manage connection status https://github.com/redhat-developer/app-services-ui-components/blob/a717e300f8cb2a2c3c7df45d1cdc5d4f7b5162d7/src/Kafka/Settings/types.ts#L1. Do you think we should create enum type?

ajaypratap003: we can use this `SettingsStatus` type which is already there. No need to create enum type.

ajaypratap003 (approved):
LGTM

src/Kafka/Settings/Settings.tsx
HemaHG: yes, I am using that only https://github.com/redhat-developer/app-services-ui-components/blob/a717e300f8cb2a2c3c7df45d1cdc5d4f7b5162d7/src/Kafka/Settings/Settings.tsx#L20

PR #470, opened 2022-07-18
chore: use description list to display the header content of consumer group drawer

HemaHG:

> Use the description list to display the content of  Active members, Partitions with lag, and State in the consumer group drawer instead of handling the layout content ourselves.

image

Update the UI to use the description list as per the issue https://github.com/patternfly/patternfly/issues/4699

> Fixes https://issues.redhat.com/browse/MGDSTRM-7703

**Verification steps** 
https://619e1921b3e38b003afeb2be-wxhhbasgoc.chromatic.com/?path=/story/kafka-consumergroups-components-consumergroupdrawer--consumer-group-drawer-details


riccardo-forina (commented):
LGTM, but I noticed testing the component on the XS breakpoint that the table is missing the column labels, so on small screens, the table doesn't work very well.

HemaHG:
I am using the responsive table for creating the table  https://619e1921b3e38b003afeb2be-prtsfroojb.chromatic.com/?path=/story/kafka-consumergroups-components-consumergroupdetails--consumer-group-details-at-kafka-level table and wrapped that table inside the Drawer components https://619e1921b3e38b003afeb2be-prtsfroojb.chromatic.com/?path=/story/kafka-consumergroups-components-consumergroupdrawer--consumer-group-drawer-details, now some columns are not visible, @riccardo-forina do you have any suggestion on how to fix this

riccardo-forina:
There is probably not enough space to show all the columns at the same time.

I don't think the responsive table is a good fit for this, as it's meant to be used in conjunction with the drawer to show the data that gets hidden.

You can look into its implementation to see how to pass the labels to the columns to get the right mobile view.

src/Kafka/ConsumerGroups/components/ConsumerGroupByTopic.tsx
```diff
@@ -100,17 +104,17 @@ export const ConsumerGroupByTopic: FunctionComponent<
...
                   {memberId ? (
                     groupId + "\n" + memberId
                   ) : (
                     <i>{t("consumerGroup.unassigned")}</i>
                   )}{" "}
                 </Td>
-                <Td key={columnNames.current_offset}>{offset}</Td>
-                <Td key={columnNames.log_end_offset}>{logEndOffset}</Td>
-                <Td key={columnNames.offset_lag}>{lag}</Td>
+                <Td dataLabel={columnNames.current_offset}>{offset}</Td>
```
HemaHG, line 115: using `dataLabel` instead of `key` fixed the small screen view issue

riccardo-forina (approved):

PR #495, opened 2022-08-05
feat: add a way to set the ouia id for the action cell in ResponsiveTable

gashcrumb:

This change adds a customizer function to ResponsiveTable so a consumer can
customize the test ID attribute for the actions kebab.  It also updates
the default test ID used so that rows can be selected by index.

I've left the return type of the customizer function constrained to just what I need, but maybe there's a better option.  Also I'm meh on the function name, so any suggestions are welcome :-)


riccardo-forina (commented):
Thanks for this! My only suggestion is, as the name of the callback is `setActionCellProps`, it implies it can be used to set any of the allowed attributes through it. In reality, it allows only for the test id to be set, which is absolutely fine, but then I'd rename it to be `setCellOuiaId` and simplify the return type

gashcrumb:
Yup updated that, agreed @riccardo-forina that looks better.

riccardo-forina (approved):

PR #496, opened 2022-08-05
fix: show the billing option in the create Kafka dialog details card …

riccardo-forina:
…only if the user can't provide an option


jgiardino (approved):
I checked the storybook and the billing info displays as expected.

PR #498, opened 2022-08-05
fix: make the billing popover show a title, and make the link not external

riccardo-forina:

image

Added the title to the popover, and made the link not external.


jgiardino (approved):
LGTM!

PR #499, opened 2022-08-05
fix: add ouiaid for terms and conditions modal

ajaypratap003:

Add ouiaId for terms and conditions modal

https://issues.redhat.com/browse/MGDX-173 

**Verification steps** 
See ouiaId in storybook component


ajaypratap003:
FYI, @jgiardino separate  ouia-id is not possible for icon-close in PF modal.

jgiardino:
Thanks, @ajaypratap003! Assuming we also display T&C for Service Registry, is it possible to have the OUIA attributes reflect that?

When I update the service from "Kafka" to "Service Registry", I still see the original Kafka OUIA attributes.  

image

ajaypratap003:
> Thanks, @ajaypratap003! Assuming we also display T&C for Service Registry, is it possible to have the OUIA attributes reflect that?
> 
> When I update the service from "Kafka" to "Service Registry", I still see the original Kafka OUIA attributes.
> 
> image

Yes, I need some changes for handling OUIAId for kafka and service registry . It should be pass as props. 
Please provide OUIAID  for service registry.

src/shared/TermsAndConditionModal/TermsAndConditionModal.tsx
```diff
@@ -8,21 +8,37 @@ import {
...
+  ouiaIdButtonViewTerms?: string;
+  ouiaIdButtonCancel?: string;
+};
+
 export type TermsAndConditionModalProps = {
   onClickViewTermsConditions: () => void;
   onCancel: () => void;
   isModalOpen: boolean;
   serviceName: string;
+  ouiaIds?: OuiaIdModal;
```
riccardo-forina, line 22: It's better to avoid passing objects as props. If the object isn't static, it will make the component re-render every time the parent component does.

HemaHG (approved):
LGTM

nwanders:
Sorry for the delay. These changes look good to me. 
I'll look into requesting an update to PF for the close action.

PR #503, opened 2022-08-10
feat: KafkaInstances components to render the list of Kafka instances

riccardo-forina:
This is a full view component, it's meant to be used as the direct child of a `Page` component as it provides all the expected visuals for this page.

Changes:
* promotes the `SearchInput` component we already had to be a shared one that can be reused. It implements the latest design guidelines (eg the arrow icon in the button instead of the magnifying glass) while keeping support for validation of the input (something that the upstream [Search input component](https://www.patternfly.org/v4/components/search-input) doesn't support yet);
* new hook [`useKafkaLabels`](https://github.com/redhat-developer/app-services-ui-components/pull/503/files#diff-4f42e2524d17cc9063c8c11c5a6362c6200d86135eb05521b42a6a9d25009091): it provides the labels relative to the shared parts of the Kafka service, like field names;
* sorting support for the `ResponsiveTable` through the `isColumnSortable` prop. The callback will be called on every column; an `undefined` response will make the column not sortable;
* fixes the `onClick` event on rows in the `ResponsiveTable`, to properly detect if the click happened inside a clickable element (like a button or link) or on the row itself;
* the `TableView` component doesn't support generic toolbars anymore. The toolbar supported is one composed by a combination of the following:
  * a filter field (single, like a search input, or multiple, with a dropdown to select the filter value)
  * button actions (eg a primary call to action)
  * icon actions (eg a refresh button)
  
  This is done in order to properly support the toolbar across all viewports, as the amount of PF code to write to do that isn't trivial.
* new hooks to use the query string portion of the URL as the single source of truth for key-value information. This is useful for things like sorting of a table, or filtering of a table, that we want to register as navigation events to allow the user to go back to the previous setting simply clicking the back button on the browser. This also allows users to share an URL, or bookmark complex filters they want to reuse. An example usage of these hooks can be seen [here](https://github.com/redhat-developer/app-services-ui-components/pull/503/files#diff-cf56f49e7edf6de23067b8671290f7382a2fd60d7a2d066f65ad541106e9dc4aR70-R80)


riccardo-forina:
@gashcrumb wanna give this a check? Btw if you want to be marked as a reviewer you'll have to add yourself as a member of https://github.com/orgs/redhat-developer

HemaHG (approved):
Great work @riccardo-forina

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -318,7 +331,13 @@ export const DeletableRow: FunctionComponent<DeletableRowProps> = memo(
     return (
       <Tr
```
gashcrumb, line 332: Do you need `key` set on these?  I feel like this might be needed.

gashcrumb: The other thing is if that's the case, would it make sense to also expose a customizer function for these keys?

gashcrumb (approved):
:+1:  this is put together really well, great stuff!

riccardo-forina:
Thank you all for the review. I pushed a new set of changes aimed at handling better the experience of the empty states (the initial no data one, and the no results one). I also added a few fixes for the UX of filtering data in the table, to ensure that we always show the right loading state and elements (like, showing the skeleton and no pagination when a new filter is set).

gashcrumb (approved):
:+1: I still approve :smile: and appreciate the inline comments in the JSX mentioning what the various blocks do.

PR #504, opened 2022-08-15
chore: ACL Filter Chip Groups

Kells512:

>Updating the ACLFilterChips and making different variants of it  

**Verification steps** 

> Original Updated
Screenshot from 2022-08-15 11-27-41

Screenshot from 2022-08-08 15-54-37
> "ACLFilterChips > < Resource Type Operation Chips Demo > ”

> One Group Verison
Screenshot from 2022-08-15 11-28-16
> "ACLFilterChips > < OneGroup > < Resource Type Operation One Group > ”

> With Patternfly Chip Group
Screenshot from 2022-08-15 11-29-00
> "ACLFilterChips > < PFChip > < Stander PF Chip Group > ”


src/ProofOfConcepts/ACLFilterChips/OneGroup.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Toolbar from "./toolbarOneGroup";
+import Chips from "./optionOneGroup";
+
+export const OneGroupChips: VoidFunctionComponent = () => {
+  return (
+    <React.Fragment>
+      {/* Mock toolbar with options pre-checked*/}
+      <Toolbar />
+      <br></br>
+      &emsp;
```
jgiardino, line 12: I don't think we use `<br>` elements when building UI. Usually, any spacing needed is handled with CSS. PF usually provides the spacing we need—a component will define space it needs inside the component, and a layout will define the space between components. Ideally, PF options for modifying spacing are what we use, but sometimes there are designs that require some customization.

For this layout, I think the Toolbar component provides what's needed to handle layout and spacing. There is this example showing a toolbar with filters that you should be able to follow and use here: https://www.patternfly.org/v4/components/toolbar#with-filters

src/ProofOfConcepts/ACLFilterChips/TrasactionalIdLabel.js
```diff
@@ -10,9 +11,12 @@ class TransactionalIdLabel extends React.Component {
     return (
       <div class="row">
         <div class="child">
-          <span class="pf-c-label mas-m-solid smaller-label pf-m-orange ">
+          <Label
+            className="pf-c-label mas-m-solid smaller-label pf-m-orange"
+            isCompact
+          >
             <span class="pf-c-label__content">TI</span>
```
jgiardino, line 18: When you replaced the first span with the `<Label>` you can also remove this span. This extra span is provided by the `<Label>` component for you.

Kells512: I applied both of the changes you mentioned :)

src/ProofOfConcepts/ACLFilterChips/ConsumerGroupLabel.js
```diff
@@ -10,9 +11,12 @@ class ConsumerGroupLabel extends React.Component {
     return (
       <div class="row">
         <div class="child">
-          <span class="pf-c-label mas-m-solid smaller-label pf-m-green ">
+          <Label
+            className="pf-c-label mas-m-solid smaller-label pf-m-green"
+            isCompact
```
jgiardino, line 16: I didn't get a chance to note some comments about this component in my last review, but there are a few items that would be good to address here.

1. While you can apply classes to PF components this way, any PF classes should be specified using the props defined for the component, and only custom classes that we create to extend the component would be defined with `className`.
2. Any custom classes we create should ideally follow a similar naming convention as the PF class names. I think [HTML training module](https://www.patternfly.org/v4/training/html) for overriding variables explains the BEM naming scheme. When we create class names, instead of using the "pf-" namespace, we would just use "mas-" so that we can easily identify which classes we've created. 
3. When we create custom modifier classes, it's good to scope them to a specific concern. E.g. "-compact" makes the component smaller, "-green" makes it green. 

Related to points 2 and 3 above, it would be good to address the `smaller-label` class and styles associated with it. This class doesn't align with other class names, but it also applies more styles than the name implies. We probably don't need a "smaller" class if we can use the PF modifier class that comes from the `isCompact`. But this needs some review to identify any other styles that were included for `smaller-label` that might still be needed, and find/define a class name for them to live under (E.g. maybe some belong under `mas-m-solid`).

Kells512: I think I fixed it in accordance to those points

Kells512:
@jgiardino I made some changes to class names and what was in them, I left alone some I wasn't sure whether to change or not

src/ProofOfConcepts/ACLFilterChips/ConsumerGroupLabel.js
jgiardino: Looking at this again, it looks like you missed converting the `pf-m-[color]` modifier classes to using the react component `color` prop. These modifier classes are PF classes, and we'll want to continue using classes that come from PF rather than recreate them as `mas-m-[color]` classes. 

You can see an example of the `color` prop in the first example on this page: https://www.patternfly.org/v4/components/label

src/ProofOfConcepts/ACLFilterChips/customStyles.css
```diff
@@ -1,38 +1,23 @@
...
-      ) * 0.85
-  );
-  min-height: calc(
-    (
-        (var(--pf-c-label--FontSize) * var(--pf-global--LineHeight--md)) +
-          (var(--pf-c-label--PaddingTop) + var(--pf-c-label--PaddingBottom))
-      ) * 0.85
-  );
   display: inline-flex;
   justify-content: center;
```
jgiardino, line 5: I'm not sure that these styles are actually needed. I think they're already included as part of the Label component. If that's the case, we should remove them.

src/ProofOfConcepts/ACLFilterChips/customStyles.css
```diff
@@ -49,32 +34,31 @@
...
+
+:root .pf-c-chip-group__close {
+  display: flex;
+  align-self: center;
 }
 
-.pf-c-chip-group.bar {
+.pf-c-chip-group.mas-m-bar {
   --pf-c-chip-group--m-category--BackgroundColor: darkgray;
 }
```
jgiardino, line 64: I'm not able to find anything that uses this class. I think it was part of the original POC where we were exploring different possible layout options. Can you check if this is actually needed, and remove if not?

src/ProofOfConcepts/ACLFilterChips/customStyles.css
```diff
@@ -1,38 +1,23 @@
-.pf-c-label.mas-m-solid.smaller-label {
+.pf-c-label.mas-m-solid.mas-m-text {
```
jgiardino, line 1: I'm not sure this extra `mas-m-text` class name is necessary. The styles defined here are what I would expect as part of a solid modifier class (e.g. transparent border and white text). This might be something to look into and see if `mas-m-text` is actually needed or not.

Kells512: It looks like the transparent border and white text are needed, but I took out everything else in it as it seems okay to.

jgiardino: I agree that the styles defined for this rule are needed. I'm not sure that we need both classes `mas-m-solid` and `mas-m-text`. If you remove `mas-m-text` from this, will the styles still be applied as expected? The class name `mas-m-solid` describes the styles that we're defining here (i.e. to create a label with a solid fill, we need to make the border transparent and the text white).

Kells512: when mas-m-text is removed, the text and border don't appear as we would like: 
**just "mas-m-text" removed**
Screenshot from 2022-08-31 14-24-41

and when the mas-m-solid is removed the text and border are also affected, and the background color of the button too:
**just "mas-m-solid" removed**
Screenshot from 2022-08-31 14-30-44

jgiardino: I think the issue is with the specificity, where the styles defined for a PF rule like `.pf-c-label.pf-m-blue` are being applied over what we're trying to define for `.pf-c-label.mas-m-solid`. 

That's not what I would expect, I would expect that whatever custom classes we define would override something with equal specificity in the PF styles. 

@dlabaj - Do you have any suggestions on how to add a custom class that doesn't get overridden by PF styles of equal specificity? Is this more related to how we're pulling in styles?

src/Kafka/ManageKafkaPermissions/components/ResourceTypeLabel.css
```diff
@@ -1,7 +1,7 @@
 .pf-c-label.mas-m-solid {
-  --pf-c-label__content--before--BorderColor: transparent;
+  --pf-c-label__content--before--BorderColor: transparent !important;
```
Kells512, line 2: @jgiardino One way of fixing the "mas-m-solid" issue I found was adding "!important" to the necessary lines in the original class in "ResourceTypeLabel.css", but not sure if this okay way to solve it.

src/ProofOfConcepts/ACLFilterChips/customStyles.css
dlabaj: It has to do with the import order.  You need to  import your css in your js files last.  Then it works correctly.

src/Kafka/ManageKafkaPermissions/components/ResourceTypeLabel.css
dlabaj, line 2: remove this important.

src/Kafka/ManageKafkaPermissions/components/ResourceTypeLabel.css
```diff
@@ -1,7 +1,7 @@
...
   --pf-c-label--BackgroundColor: var(--pf-global--palette--black-600);
-  --pf-c-label__content--Color: var(--pf-global--palette--white);
+  --pf-c-label__content--Color: var(--pf-global--palette--white) !important;
```
dlabaj, line 4: Remove this important

src/ProofOfConcepts/ACLFilterChips/KafkaInstanceLabel.js
```diff
@@ -1,5 +1,6 @@
 import "./customStyles.css";
 import React from "react";
+import { Label } from "@patternfly/react-core";
```
dlabaj, line 3: You need to move import "./customStyles.css"; after you import the Label in all the .js files.

dlabaj (changes requested):

src/ProofOfConcepts/ACLFilterChips/KafkaInstanceLabel.js
Kells512: I applied this and removed the "!important" from "ResourceTypeLabel.css",, but "mas-m-solid" still doesn't seem to apply

jgiardino:
Looking at the components (vs the code), they look good to me!

jgiardino (approved):
LGTM

PR #506, opened 2022-08-18
fix: specify the pagination variants

gashcrumb:

This change ensures that the pagination navigation is available in
smaller screen widths by specifying the appropriate variant to the
pagination control.  It also removes the toolbar wrapping the pagination
component as it prevents the pagination control from appearing correctly
in narrow widths.

The bug in action (yes I like loud backgrounds):

image

with this fix applied in storybook:

image


riccardo-forina (approved):
Thank you!

PR #507, opened 2022-08-19
chore: Add data-testid for select and dropdown component in kafka message browser

HemaHG:

Add data-testid for select and dropdown component in kafka message browser

> Add data-testid as per the QE request

### DIFF
--- a/src/Kafka/KafkaMessageBrowser/components/LimitSelector.tsx
+++ b/src/Kafka/KafkaMessageBrowser/components/LimitSelector.tsx
@@ -40,6 +40,7 @@ export const LimitSelector: VoidFunctionComponent<LimitSelectorProps> = ({
           isOpen={isOpen}
           isDisabled={isDisabled}
           onSelect={() => setIsOpen(false)}
+          data-testid={"limit-selector"}
         >
           {[10, 20, 50].map((value, idx) => (
             <SelectOption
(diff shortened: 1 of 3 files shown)


src/Kafka/KafkaMessageBrowser/components/LimitSelector.tsx
```diff
@@ -40,6 +40,7 @@ export const LimitSelector: VoidFunctionComponent<LimitSelectorProps> = ({
           isOpen={isOpen}
           isDisabled={isDisabled}
           onSelect={() => setIsOpen(false)}
+          data-testid={"limit-selector"}
```
mmusil, line 43: For some reason, this ID is not on the div that I would need it to be, but on <ul> element when you expand the select component.
image

src/Kafka/KafkaMessageBrowser/components/PartitionSelector.tsx
```diff
@@ -106,6 +106,7 @@ export const PartitionSelector: VoidFunctionComponent<
           isDisabled={isDisabled}
           placeholderText={t("partition_placeholder")}
           onClear={() => onChange(undefined)}
+          data-testid={"partition-selector"}
```
mmusil, line 109: Same as above, but I will test it if I can make it work in the test suite.

src/Kafka/KafkaMessageBrowser/components/LimitSelector.tsx
riccardo-forina: That's unfortunately how the PF select handles the attribute, we can't do much about it I think.

mmusil: In that case, I will use `/parent::div` in the selector then... Thanks!

mmusil (approved):

PR #531, opened 2022-09-06
chore: Refresh button for Topics and Consumer Groups UI 

Kells512:

> A refresh button that can be used for Consumer Groups and Topics

**Verification steps** 

> **Hoover**
> < PROOFOFCONCEPTS > < RefreshButton > < Example1 >
>Screenshot from 2022-09-19 13-35-54

> **Text next to button**
> < PROOFOFCONCEPTS > < RefreshButton > < Example2 >
> Screenshot from 2022-09-19 13-36-10

> **Text to far right*
> < PROOFOFCONCEPTS > < RefreshButton > < Example3  >
> Screenshot from 2022-09-19 13-36-20


jgiardino:
We chatted about showing a few of variations of this component, to have some options to see what works best where we need it. These are some variants that I think would be good to try. 

- **Variant you have now** - For the tooltip to support screen readers, we need text that describes the action. The "Last updated..." text is more like a description, rather than a label. In past discussions about this action for message browser, the suggestion was that this is a "Reload" action. Let's try a tooltip that is something like this:
  - > Reload contents
      > Last updated 8 minutes ago
- **Variant similar to what's in the Metrics example** - I think we will need this regardless, assuming we would want a shared component that we can use in the toolbar on the metrics page.
  - You will still need a tooltip for the button, but in this case it should just say "Reload contents"
- **Variant similar to what's in the Metrics example, but with text immediately to the right of the action** - In the context of the table toolbar, this might be the best option.
  - Same comment as above - You will still need a tooltip for the button, but in this case it should just say "Reload contents"

Eventually, we would want to try these variants out in different examples, like the table view toolbar and the metrics page that we have now. I'm not sure if this is something we should wait for @riccardo-forina to come back so that he can provide suggestions on this. But if this is something you are interested in exploring with this PR then I think that's fine, too, and we can just sync with Riccardo when he's back on next steps before merging.

Kells512:
Hey @jgiardino, I think both this pr and [504](https://github.com/redhat-developer/app-services-ui-components/pull/504) are done now, just would like one last look over whenever you can :)

jgiardino:
Your updates look really good! There are a couple of issues I noticed which are noted below.

A tooltip displays for the action to describe what the action does, but when the spinner displays, I wouldn't expect the same tooltip to display for the spinner.
image

The layout overall looks really good. The one place where it seems slightly off is when isRefreshing=true, the spinner sits lower than I would expect it to. Ideally this would be centered in the row the same way the button is. I poked around a bit but didn't see any obvious reason why this isn't centered. I did notice that the class `pf-m-in-progress` is missing from the progress button, but that seems to only affect the color of the spinner. And when I look at the [PF example of this button](https://www.patternfly.org/v4/components/button#progress-indicators), the PF example looks centered as expected, so I'm not sure what's causing the layout to be off here. 
| isRefreshing = false | isRefreshing = true |
| -- | -- |
| image | image |
| image  | image |

Kells512:
some issues with fixing tests, so made a new [pr](https://github.com/redhat-developer/app-services-ui-components/pull/560) which is fine now.

PR #532, opened 2022-09-07
chore: re-export react-i18next 

kelvah:

We are currently required to install `react-i18next` as a dependency of our MAS UI projects in order to translate content.

This will frequently produce two different versions of the same package in the dependency tree, like:

```
├─┬ @rhoas/app-services-ui-components@1.40.6
│ └── react-i18next@11.17.3
└── react-i18next@11.18.5
```

`I18nProvider` and `useTranslation` will then come from different versions. The result is unit tests will start failing, they will not be able to translate strings anymore.
 
One possible solution is to re-export `react-i18next` from `app-services-ui-components`.

This way it will live only here. Other UI projects will not have to include `react-i18next` as a dependency. Instead they will import everything from `app-services-ui-components`, both i18n providers and translation functions.

There are other ways to fix this issue but I guess re-exporting is the simpler one.

It's been a while since we last discussed this issue. Let me know what you think or if there are different plans about how to sort things out.

### DIFF
--- a/src/I18n/I18nProvider.tsx
+++ b/src/I18n/I18nProvider.tsx
@@ -22,3 +22,5 @@ export const I18nProvider: FunctionComponent<I18nProviderProps> = ({
     {children}
   </I18nextProvider>
 );
+
+export * from "react-i18next";
--- a/README.md
+++ b/README.md
@@ -60,10 +60,6 @@ module.exports = () => {
             singleton: true,
             requiredVersion: peerDependencies['react-dom'],
           },
-          'react-i18next': {
-            singleton: true,
-            requiredVersion: peerDependencies['react-i18next'],
-          },
           'react-router-dom': {
             singleton: true,
             requiredVersion: peerDependencies['react-router-dom'],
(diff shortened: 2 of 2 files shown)


src/I18n/I18nProvider.tsx
```diff
@@ -22,3 +22,5 @@ export const I18nProvider: FunctionComponent<I18nProviderProps> = ({
     {children}
   </I18nextProvider>
 );
+
+export * from "react-i18next";
```
jomarko, line 26: Thank you @kelvah . I have a comment to this change. Is it save to export `all` from an existing library? 

Shouldn't we at least guide users in documentation to not install `react-i18next` dependency?

or shoudln't  we maybe export just `useTranslation` ?

or maybe to not introduce another method with same name maybe we shouldn export just some wrapper like?

```
export appServicesUseTranslation = useTranslation();
```

kelvah: > Thank you @kelvah . I have a comment to this change. Is it save to export `all` from an existing library?
> 
> Shouldn't we at least guide users in documentation to not install `react-i18next` dependency?

I think I added this to the `README`. If it's not clear enough we can of course make it more explicit.

> or shoudln't we maybe export just `useTranslation` ?
> 
> or maybe to not introduce another method with same name maybe we shouldn export just some wrapper like?
> 
> ```
> export appServicesUseTranslation = useTranslation();
> ```

I was re-exporting only `useTranslation` initially. Then I've realized we also need the [`Trans` component](https://react.i18next.com/latest/trans-component) for smart events. So I thought it was easier to just re-export the whole library.

If someone needs something that's not exported or if a new relevant feature will be added in future releases of `react-i18next`, we'll have to make new `ui-components` releases every time. Maintaining a sub list of exports looks a bit annoying.

riccardo-forina (approved):
Sorry, I was out. I think this is a good way to ease the pain of managing deps, let's see how it works out there.

jomarko:
@riccardo-forina is there any estimation when this PR can be merged please?

riccardo-forina:
Done and marked as a feat so to trigger a release

PR #537, opened 2022-09-12
chore: Fix semantic release failing

suyash-naithani:
Semantic release is failing on the repo. This PR aims to fix that


riccardo-forina:
I think we can close this

PR #544, opened 2022-09-15
chore: Sync package and package-lock

suyash-naithani:
Include postcss upgraded version.
Upgrade storybook versions


HemaHG:
@suyash-naithani we need to match the dependencies with the other repository dependency https://github.com/redhat-developer/app-services-ui/pull/631#issuecomment-1245648531, so that it should show the error

suyash-naithani:
> @suyash-naithani we need to match the dependencies with the other repository dependency [redhat-developer/app-services-ui#631 (comment)](https://github.com/redhat-developer/app-services-ui/pull/631#issuecomment-1245648531), so that while updating components version in other repository should not show the error

Yes, but we first need to make sure that storybook and it's test in this PR pass, which is not currently the case

PR #546, opened 2022-09-19
feat: Topic table

HemaHG:

> Use the responsive table for the topics list view

fixes https://issues.redhat.com/browse/MGDSTRM-9359

Link to story - https://619e1921b3e38b003afeb2be-lajvitpdte.chromatic.com/?path=/story/kafka-kafkatopics-kafkatopics--kafka-topics

### DIFF
--- a/locales/en/topic.json
+++ b/locales/en/topic.json
@@ -1,4 +1,20 @@
 {
   "create_topic": "Create topic",
-  "show_all_options": "Show all available options"
+  "empty_topics_body": "To get started, create a topic.",
+  "empty_topics_title": "No topics",
+  "input_field_invalid_message": "Valid characters include letters from a to z, numbers from 0 to 9, and hyphens ( - ).",
+  "partitions": "$t(create-topic:partitions)",
+  "retention_size": "Retention size",
+  "retention_time": "Retention time",
+  "show_all_options": "Show all available options",
+  "table": {
+    "actions": {
+      "delete": "Delete topic",
+      "edit": "Edit topic configuration"
+    }
+  },
+  "topic_list_table": "Topic list table",
+  "topic_name": "Name",
+  "topic_search": "Search topics button",
+  "topic_search_input": "Topic search input"
 }
--- a/locales/en/common.json
+++ b/locales/en/common.json
@@ -34,6 +34,8 @@
   "distance-to-now-ago": "{{date}} ago",
   "duration": "Duration",
   "duration_value": "2 months",
+  "empty_state_no_results_body": "No results match the filter criteria. Remove filter or clear all filters to show results.",
+  "empty_state_no_results_title": "No results found",
   "failed": "Failed",
   "finish": "Finish",
   "form_invalid_alert": "Please address the errors in your form to proceed",
--- a/src/Kafka/CreateKafkaTopic/CreateTopicPage.stories.tsx
+++ b/src/Kafka/CreateKafkaTopic/CreateTopicPage.stories.tsx
@@ -1,7 +1,7 @@
 import type { ComponentStory, ComponentMeta } from "@storybook/react";
 import { userEvent, within } from "@storybook/testing-library";
 import { fakeApi } from "../../shared/storiesHelpers";
-import { RetentionSizeUnits, RetentionTimeUnits } from "./types";
+import { RetentionSizeUnits, RetentionTimeUnits } from "../types";
 import { CreateTopicPage } from "./CreateTopicPage";
 import { constantValues } from "./components/storiesHelpers";
 
--- a/src/Kafka/CreateKafkaTopic/components/CoreConfiguration.tsx
+++ b/src/Kafka/CreateKafkaTopic/components/CoreConfiguration.tsx
@@ -19,9 +19,8 @@ import { CustomRetentionSize } from "./CustomRetentionSize";
 import {
   retentionTimeSelectOptions,
   retentionSizeSelectOptions,
-  RetentionTimeUnits,
-  RetentionSizeUnits,
 } from "../types";
+import { RetentionSizeUnits, RetentionTimeUnits } from "../../types";
 import { useValidateTopic } from "../types";
 import type { NewTopic } from "../types";
 
--- a/src/Kafka/CreateKafkaTopic/components/CustomRetentionMessage.tsx
+++ b/src/Kafka/CreateKafkaTopic/components/CustomRetentionMessage.tsx
@@ -8,7 +8,8 @@ import {
   SelectVariant,
 } from "@patternfly/react-core";
 import type { SelectProps, NumberInputProps } from "@patternfly/react-core";
-import type { NewTopic, RetentionTimeUnits, SelectOptions } from "../types";
+import type { NewTopic, SelectOptions } from "../types";
+import type { RetentionTimeUnits } from "../../types";
 
 export type CustomRetentionMessageProps = NumberInputProps &
   SelectProps & {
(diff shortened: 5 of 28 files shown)


src/Kafka/KafkaTopic/types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export type KafkaTopicField = keyof KafkaTopic;
+
+export const unitsToBytes = {
+  [RetentionSizeUnits.BYTE]: 1,
+  [RetentionSizeUnits.KIBIBYTE]: 1024,
+  [RetentionSizeUnits.MEBIBYTE]: 1048576,
+  [RetentionSizeUnits.GIBIBYTE]: 1073741824,
+  [RetentionSizeUnits.TEBIBYTE]: 1.0995116e12,
+};
```
riccardo-forina, line 18: For the metrics we use an [external lib](https://github.com/redhat-developer/app-services-ui-components/blob/b97ccae1133e0ccec1c1f2ece120e8aaf8f441af/src/Kafka/Metrics/components/utils.ts#L7-L9) to convert bytes. In general, for solved problems like these, I'd prefer relying on something external (end tested) than writing some custom code ourselves.

No need to address this in this PR, but I'd make a note to streamline all these conversions. It's clear how it's a shared concern across components, so we should have some shared component/utility to deal with it.

src/Kafka/KafkaTopic/KafkaTopics.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export type KafkaTopicsProps<T extends KafkaTopic> = {
+  getTopics: (
+    page: number,
+    perPage: number
+  ) => Promise<{
+    topics: T[];
+    count: number;
+  }>;
+};
```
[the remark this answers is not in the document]
HemaHG: apologies @riccardo-forina, I am still working on it and didn't notice that have pushed something that's still in WIP.

src/Kafka/KafkaTopic/types.ts
HemaHG: yeah sure, will make use of the external lib

src/Kafka/KafkaTopic/components/EmptyStateNoResults.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <EmptyState variant={EmptyStateVariant.large}>
+      <EmptyStateIcon icon={SearchIcon} />
+      <Title headingLevel="h3" size="lg">
+        {t("common:empty_state_no_results_title")}
+      </Title>
+      <EmptyStateBody>{t("common:empty_state_no_results_body")}</EmptyStateBody>
+    </EmptyState>
+  );
+};
```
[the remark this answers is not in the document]
HemaHG: moved to the shared folder, as it's used by other components like the consumer table and ACLS table

HemaHG:
@riccardo-forina, the existing search filter in the topic table contains search input with only chips and the value in the chips will be overwritten if we search, but the new search filter contains the chip group, can I make use of the search input component if we only have a search filter or can I makes changes in the chip filter so that it only takes a string input instead of string

HemaHG:
@jgiardino gave heads-up to use the new search filter

locales/en/topic.json
```diff
@@ -1,4 +1,20 @@
...
+  "empty_topics_title": "No topics yet",
+  "input_field_invalid_message": "Valid characters include letters from a to z, numbers from 0 to 9, and hyphens ( - ).",
+  "partitions": "$t(create-topic:partitions)",
+  "retention_size": "Retention size",
+  "retention_time": "Retention time",
+  "show_all_options": "Show all available options",
+  "table": {
+    "actions": {
+      "delete": "Delete",
+      "edit": "Edit"
```
jgiardino, line 13: As part of refactoring the Kafka table view, we updated the labels in the action menu based on usability test findings. Can we do the same for the actions in this menu?

The updates would be:
* Edit 👉 Edit topic configuration
* Delete 👉 Delete topic

locales/en/topic.json
```diff
@@ -1,4 +1,20 @@
 {
   "create_topic": "Create topic",
-  "show_all_options": "Show all available options"
+  "empty_topics_body": "Create a topic by clicking the button below to get started.",
+  "empty_topics_title": "No topics yet",
```
jgiardino, line 4: This was one of the first UIs we implemented, and the contents have evolved slightly since this empty state was included. Is it possible to also update these contents to be consistent with others:
```
  "empty_topics_body": "To get started, create a topic.",
  "empty_topics_title": "No topics",
```

jgiardino:
These updates look really good! I posted a couple of text string updates inline. The only other thing I noticed is related to the filter. I'm not sure if this is addressable in this PR or how to approach this, but wanted to capture it here to figure out next steps.

In the Kafka instances table, we have many filters. One of the indicators that this UI is for filtering is the funnel/filter icon that displays in the select menu, where you select what property to filter on.

image

In the filter for topics, this select menu is not included, and therefore this additional visual indication is not included.

image

I think the recommendation based on [what I asked in slack](https://patternfly.slack.com/archives/CGR204RM2/p1666976896041989) is that the text input should always have a search icon, whether it's preceded by a select menu or not.

This is consistent with what's shown on the filter guidelines: 
https://www.patternfly.org/v4/guidelines/filters#text-entry-filters
https://www.patternfly.org/v4/guidelines/filters#attribute-value-filter

I think this means we just need to add a 🔍 icon to the text input that displays in the filter bar, and the funnel/filter icon only displays on the select toggle. But I think this is likely a separate update from this PR. 🤷‍♀️

HemaHG:
sure @jgiardino, I will make [this](https://github.com/redhat-developer/app-services-ui-components/pull/546#issuecomment-1295365822) updates In another PR

jgiardino:
I see the updates to the text strings were made. One other thing I noticed is the alignment of the topic names

This is happening as a result of not using the inline modifier class on the link that displays in that column. There should be a prop for the Button component you can use to fix that alignment issue.

image

HemaHG:
@jgiardino, I have incorporated all the review comments, I have also addressed your review comment https://github.com/redhat-developer/app-services-ui-components/pull/546#issuecomment-1295365822 to include the search icon in the search input field

jgiardino (approved):
Thanks for making all the updates I suggested.

jgiardino (changes requested):
I think this is an update from this PR. When I look at [Type Error For Invalid](https://619e1921b3e38b003afeb2be-lajvitpdte.chromatic.com/?path=/story/shared-tabletoolbar-searchinput--type-error-for-invalid), I see the following:

image

jgiardino:
This last comment would be a nice-to-have if it's simple, but since I didn't catch it the first time, I don't want to hold this PR up if it's a lot of effort.

One of the patterns we're trying to standardize on is to have the Delete action always be the last item in a menu. If it's possible, could you swap the order that these menu items display in?

image

HemaHG:
swapped the order of action items https://619e1921b3e38b003afeb2be-jvkoijhyux.chromatic.com/?path=/story/kafka-kafkatopics-kafkatopics--kafka-topics and removed the search filter icon, I will handle that in a separate PR

riccardo-forina (approved):
LGTM. As a next enhancement I'd like to see us start using `convert` for the bytes, as it's now a supported data type https://convert.js.org/modules#convert

jgiardino (approved):
Thanks for making the updates!

PR #550, opened 2022-09-21
fix: Fix issues of #525

jomarko:
The autogenerated PR #525 is failing. It needs some adjustments to be passing CI checks. Providing adjustments via this PR.

Changes are split into two commits:
- 'children' in props
- some explicit typing of callback parameters

**Verification steps** 


riccardo-forina:
Thank you, but we don't plan on updating to the latest React for a while, until consoledot does

jomarko:
@riccardo-forina ok, shouldn't be then closed also #525 to not bring confusion. because right now such update is planed.

riccardo-forina:
That's correct, I did close that before replying to you

PR #552, opened 2022-09-22
fix: upgrade ui-shared and other peer dependency

ajaypratap003:
fix: upgrade ui-shared and other peer dependency


riccardo-forina (changes requested):
Peer dependencies can't be pinned like that, we need to accept a range, like it is now

riccardo-forina (approved):

PR #563, opened 2022-10-13
chore: Allow user to type custom time in timestamp filter

suyash-naithani:
Signed-off-by: suyash-naithani <suyash.naithani@gmail.com>

> Allow user to input a custom time duration

https://issues.redhat.com/browse/MGDSTRM-8723


src/Kafka/KafkaMessageBrowser/components/DateTimePicker.tsx
```diff
@@ -23,17 +25,30 @@ export const DateTimePicker: VoidFunctionComponent<DateTimePickerProps> = ({
...
       let newDate = date;
       if (hour !== undefined) {
         newDate = setHours(newDate, hour);
       }
       if (minute !== undefined) {
         newDate = setMinutes(newDate, minute);
       }
+
       onChange(formatISO(newDate));
+      setIsTimeSelected(true);
```
riccardo-forina, line 51: When is this set back to `false`?

suyash-naithani: We would not want to set this to false, we only want this to be false when the user has not interacted with the time picker component so that he can see the placeholder rather than the default time 0:00AM for the date he has selected. Once the user starts typing or selects a value, we will no longer need the placeholder.
We use this to prevent 0:00AM to show up after we select a date as that is an invalid format. We want the initial value to be a placeholder

riccardo-forina (approved):
I still think using `onBlur` would have been a better option, but the widget is usable on Storybook so I'll give it a pass.

PR #564, opened 2022-10-14
chore: Fix Epoch timestamp filter on message browser

suyash-naithani:
Signed-off-by: suyash-naithani <suyash.naithani@gmail.com>

> This PR fixes the following epoch timestamp issue
1. The input field was behaving strange on backspace by not allowing the user to delete the value in the field
2. On increasing the value by what should be 1, the value was getting increased to a random value 

> https://issues.redhat.com/browse/MGDSTRM-8724


riccardo-forina (approved):

jgiardino (approved):
LGTM

PR #565, opened 2022-10-14
feat: Add chart for storage per broker

hemahg:

- [x]  Add filter to filter by Broker
- [x]  Add toggle filter for broker.
- [x]  Add a chart to show storage per instance
- [x]  Add a chart to show storage per broker

fixes https://issues.redhat.com/browse/MGDX-258

**Verification steps** 

https://619e1921b3e38b003afeb2be-oxzvwcabfv.chromatic.com/?path=/story/kafka-metrics--all-ready


jgiardino:
I just quickly looked at this, and it looks really good so far. Is it possible to update the examples so that we can test what the legend looks like when there are 20 partitions listed in the chart?

HemaHG:
@jgiardino @riccardo update the used disk space and partition size chart to use chart line, please have a look at story https://619e1921b3e38b003afeb2be-wncgxfjgsf.chromatic.com/?path=/story/kafka-metrics--all-ready

riccardo-forina:
Thank you. I still feel like a stacked chart here would be easier to digest... could you maybe try that? Without committing any changes, just hacking it quickly to capture a screenshot so to compare

jgiardino:
Thanks for making that update! Is it possible to include Riccardo's suggestion and also have the current line chart example so that both options are available for preview? This would let others on the team to easily review and compare the options.  But ideally, we should put this in front of users for their thoughts on how the chart displays when rendering real data, so whatever option we choose now would possibly change after this is in production.

Regarding my previous suggestion above:
> Is it possible to update the examples so that we can test what the legend looks like when there are 20 partitions listed in the chart?

I now realize that the legend _is_ updating to show 10 or 20 partitions, but it just isn't visible depending on the viewport width. In the [PF slack chat](https://patternfly.slack.com/archives/CL6EU2526/p1668438772087949), it looks like there were a couple of examples you could try here. But if these don't work out and you're stuck on this one, let me know, and I can see if we could get someone from PF dev to review the implementation and provide more suggestions. 

image

HemaHG:
@riccardo-forina , I created a new stack chart https://619e1921b3e38b003afeb2be-fobbmqmszz.chromatic.com/?path=/story/kafka-metrics-components-partitionstackchartperbroker--stack-chart, I am passing the data correctly but still, the  graph is shown only for the end time, could you please have a look

jgiardino:
Regarding this note from one of my previous comments:
> TBD if we need to add an inline info alert for the Connection attempt rate like the one we have in Client connections. Keith Wall was going to check into this

Following recent discussion with Keith, we do not have per-broker metrics for the **Connection attempt rate** chart. Therefore, we will need to include the following message in an inline info alert in the chart, similar to the one we have for **Client connections**:
> Connection attempt rate is shown for all brokers

_Screencap of the current UI preview, showing an inline alert for **Client connections** but not **Connection attempt rate**_
image

jgiardino:
@HemaHG These are a couple of things I missed in my original review and just noticed them today. Sorry for catching them so late.

1. The **Disk space size** chart use to have a limit line in the legend, but this is no longer displaying. Whenever a chart shows a limit line, there should be an entry in the legend communicating that it represents the limit.

    | What we have today | What displays for this PR |
    | -- | -- |
    | image | image |

2. The inline info alert that displays in the **Client connections chart** should only display when the user has filtered the column by a specific broker. When this inline alert is added to the **Connection attempt rate** chart, this same rule applies there, too. 
    image

hemahg:
closed in favor of https://github.com/redhat-developer/rhosak-ui/pull/77

hemahg:
closing this as it's already implemented in the rhosak-ui

PR #567, opened 2022-10-19
chore: Disable option in the kebab option when the instance is suspended

HemaHG:

> Disable changeowner and connection option in the Kafka instance kebab menu when the instance is suspended 
> Instance name as a plain text

> related to issue https://issues.redhat.com/browse/MGDSTRM-9745

**Link to story**

https://619e1921b3e38b003afeb2be-ggcjdxwohs.chromatic.com/?path=/story/kafka-kafkainstances--user-baz-can-only-delete-his-instances


riccardo-forina (commented):
I prefer to have any logic in the top `KafkaInstances` component than in the `InstancesTable` one.

Also, disabling the connection action here doesn't stop the user to see the connection tab in the drawer opening first the details one, right?

HemaHG:
if we open the connection tab from the details options connection tab will be enabled but will show an empty state
https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-kafkainstancedrawer-kafkaconnectiontabp2--connection-tab-when-kafka-instance-suspended

src/Kafka/KafkaInstances/components/InstancesTable.tsx
```diff
@@ -202,21 +202,27 @@ export const InstancesTable = <T extends KafkaInstance>({
...
+                      onClick: () => onConnection(row),
+                    }),
               },
               {
                 isSeparator: true,
               },
               {
                 title: t("table.actions.change-owner"),
-                ...(!changeOwnerEnabled
+                ...(!changeOwnerEnabled || row.status === "suspended"
```
HemaHG, line 218: @riccardo-forina, should I need to handle the logic only related to suspended in the `KafkaInstance` Component?

riccardo-forina: Yeah. Here we should have only `true` and `false`, coming from somewhere (a callback as it is here). It's in the callback that we should have the logic that checks for the suspended status.

HemaHG:
The API is still not available for suspended instances, but I have created the PR to demo this feature in the Sprint review meeting

src/Kafka/KafkaInstances/KafkaInstances.stories.tsx
```diff
@@ -67,6 +68,7 @@ export const OrgAdminCanDeleteAnyInstance = Template.bind({});
 OrgAdminCanDeleteAnyInstance.args = {
   canChangeOwner: () => true,
   canDelete: () => true,
+  instanceSuspended: () => true,
```
riccardo-forina, line 71:
```suggestion
  canOpenConnection: () => true,
```
the idea is to remove any knowledge of the _why_ something is disabled in this component. `istanceSsspended` isn't any different from `row.status === 'suspended'`

src/Kafka/KafkaInstances/components/InstancesTable.tsx
```diff
@@ -184,8 +186,10 @@ export const InstancesTable = <T extends KafkaInstance>({
         );
       }}
       renderActions={({ row, ActionsColumn }) => {
-        const changeOwnerEnabled = canChangeOwner(row);
+        const changeOwnerEnabled =
+          !canChangeOwner(row) || instanceSuspended(row);
```
riccardo-forina, line 190: Using the above suggestion, there is no need to change this. The logic should all live in `KafkaInstances`.

jgiardino:
> When the instance is suspended, the instance name appears as plain text, the text is aligned slightly to the left as shown in the below image does it look odd @jgiardino @riccardo-forina? https://619e1921b3e38b003afeb2be-vbqycokdnh.chromatic.com/?path=/story/kafka-kafkainstances--user-baz-can-only-delete-his-instances

I would expect it to be aligned with the other names. 

When looking at the example for the Deleting status (shown below), the text is still wrapped with a link, but the link is disabled. I think we would want a similar treatment here because they are functionally similar (the user cannot access the contents on the resource-level page), but I don't have a strong opinion on whether the name should be a disabled link, or just plain text.

If we display it as a disabled link, it will be gray and not black as shown in the [design document](https://marvelapp.com/prototype/2g2cb10g/screen/88526936).

If we remove the disabled link and just use plain text, then I think the alignment would be fixed by using the prop that adds the className `pf-m-inline` to the rows that do have links. Without that class, additional padding is added around the link. In the context of a table cell, that additional padding isn't needed because the table cell has it's own padding.

image

HemaHG:
created 2 versions of suspendedPopover, in the 1st version doesn't contain the external link, 2nd version contains contains external link to the doc

src/Kafka/KafkaInstances/components/InstancesTable.tsx
```diff
@@ -128,7 +130,10 @@ export const InstancesTable = <T extends KafkaInstance>({
                           {row.name}
                         </Link>
                       )}
-                      isDisabled={DeletingStatuses.includes(row["status"])}
+                      isDisabled={
+                        DeletingStatuses.includes(row["status"]) ||
```
riccardo-forina, line 134: I think we should move this logic up a level like for the other `canXYZ` callbacks. WDYT?

src/Kafka/KafkaInstances/components/InstancesTable.tsx
```diff
@@ -198,14 +204,20 @@ export const InstancesTable = <T extends KafkaInstance>({
...
               {
                 title: t("table.actions.connection"),
-                onClick: () => onConnection(row),
+                ...(!changeConnectionEnabled
+                  ? {
+                      isDisabled: true,
+                    }
+                  : {
+                      onClick: () => onConnection(row),
+                    }),
```
riccardo-forina, line 213: We have this kind of logic repeated a few times already in this component. It sounds like it's going to be a common pattern; maybe we should have a helper in the shared folder to generate these items.

jgiardino:
The additional status for resuming and suspending look good!
>what will happen to the Instance link when the instance is in resuming or suspending status

That's a good question. @riccardo-forina what would you recommend? I'm included to display the instance name the same way it displays for the creating status for both `resuming` and `suspending` (i.e. it's not a link, you can't navigate to the resource-level view). I'm not sure if there is a page we could navigate to until it's in a `ready` state. I don't think we should let them navigate to the page during the `suspending` state, since it will soon become unusable.

riccardo-forina:
Fully agree

riccardo-forina (changes requested):
Sorry spotted a problem that needs to be addressed. Edit, awesome, somehow my comment got swallowed. I'll rewrite it here.

Reading the name of the function `canHaveInstanceLink` I expect it to return `true` when an instance can have a link (`ready` state) or `false` when it can't (all other states). But then the implementation of the function is the opposite, and the function is used considering this opposite logic rather than the one that the name suggests.

Please _always_ double-check the names given to things and their behavior. Address this by:
* simplifying the implementation of the function to just check for the `ready` state, removing all those ORs
* check and update the code where the function is used so it's aligned with the corrected return value
* write tests that explicitly check for an instance to have or not have a link based on its status. Writing a Storybook story isn't enough, any new or updated behavior must be checked with explicit `expect` checks unless already covered.

src/Kafka/KafkaInstanceStatus/components/StatusLabel.stories.tsx
```diff
@@ -156,3 +162,27 @@ suspended.parameters = {
     },
   },
 };
+
+export const suspending = Template.bind({});
```
riccardo-forina, line 166: Not a huge deal but could you please do as the action suggests?

src/Kafka/KafkaInstanceStatus/components/SuspendedPopover1.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useTranslation } from "react-i18next";
+import { ExternalLink } from "../../../shared";
+import type { PopoverProps } from "@patternfly/react-core";
+
+export type SuspendedPopover1Props = {
+  children?: PopoverProps["children"];
+  reference?: RefObject<HTMLButtonElement>;
+};
+
+export const SuspendedPopover1: VoidFunctionComponent<
```
riccardo-forina, line 13: Is this component meant to be here? I don't see it used anywhere. In case we do, can we please find a better name for this? Or at least add some comment that describes why we need it?

src/Kafka/KafkaInstances/KafkaInstances.stories.tsx
```diff
@@ -16,11 +12,11 @@ export default {
   args: {
     getUrlForInstance: (row) => `/${row.id}`,
     isRowSelected: () => false,
-    canChangeOwner: (row) => row.owner === "baz-owner",
+    canChangeOwner: (row) =>
+      row.owner === "baz-owner" || row.status === "suspended",
```
riccardo-forina, line 16: I know it's a story, but can suspended instances be moved to another owner?

src/Kafka/KafkaInstanceStatus/components/StatusLabel.stories.tsx
HemaHG: done

src/Kafka/KafkaInstanceStatus/components/SuspendedPopover1.tsx
HemaHG: Initially, I created a component with an external link but then I got to know that the external link is not yet ready so created another component with not no link.

src/Kafka/KafkaInstances/KafkaInstances.stories.tsx
HemaHG: changed the story. actually, I am writing a test to check the actions items based on the status but for some reason not able to get the action button based on aria-label,

src/Kafka/KafkaInstanceStatus/components/SuspendedPopover1.tsx
riccardo-forina: `SuspendedPopoverNoSupportLink` I think would be a better name then, don't you think?

src/Kafka/KafkaInstances/KafkaInstances.stories.tsx
HemaHG: I am writing a test to check the actions items based on status, but unable to access the action button, and the test is failing with the below error https://github.com/redhat-developer/app-services-ui-components/actions/runs/3581280261/jobs/6024192336#step:4:484. @riccardo-forina do you have any idea about this

riccardo-forina: I'm not super sure what it could be, main builds fine 😕

HemaHG: I get this warning when I tried to access the action button in the instance row, I can see the same waring message locally and in stage

riccardo-forina (approved):

PR #572, opened 2022-11-03
chore: POC for updated list of Connectors

Kells512:

> Making a list of the Connectors 

**Verification steps** 

> Screenshot from 2022-11-28 11-36-29

> < PROOFOFCONCEPTS > < connectors >

>The icons on the far left are temporary till the ones in the mock-up are available


src/ProofOfConcepts/connectors/connectors.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default {
+  component: DataListSelectableRows,
+  args: {},
+} as ComponentMeta<typeof DataListSelectableRows>;
+
+const Template: ComponentStory<typeof DataListSelectableRows> = (args) => (
+  <DataListSelectableRows {...args} />
+);
+
+export const Story = Template.bind({});
```
gashcrumb, line 15: It's worth getting into the habit of avoiding generic names, often you'll find you'll want to add a second story and then you have to go back and rename this one :smile:

src/ProofOfConcepts/connectors/connectors.tsx
```diff
@@ -0,0 +1,363 @@
...
+  DataListAction,
+  Flex,
+  FlexItem,
+  Label,
+  LabelGroup,
+} from "@patternfly/react-core";
+import PlusCircleIcon from "@patternfly/react-icons/dist/esm/icons/plus-circle-icon";
+import OutlinedQuestionCircleIcon from "@patternfly/react-icons/dist/esm/icons/outlined-question-circle-icon";
+
+export const DataListSelectableRows: React.FunctionComponent = () => {
```
gashcrumb, line 17: It's a good habit to keep the filename consistent with the component name.  Maybe going forward once the components start to feel more specific you might use a name like `ConnectorTypeListItem` for example.

src/ProofOfConcepts/connectors/connectors.tsx
```diff
@@ -0,0 +1,363 @@
...
+
+  const handleInputChange = (
+    id: string,
+    _event: React.FormEvent<HTMLInputElement>
+  ) => {
+    setSelectedDataListItemId(id);
+  };
+
+  return (
+    <React.Fragment>
```
gashcrumb, line 33: Do we need this wrapping fragment?  If not it's worth removing it, even just to help reduce the indentation.

locales/en/common.json
```diff
@@ -20,6 +20,18 @@
...
+    "amazon_chip": "Amazon",
+    "amazon_queue_service": "Amazon Simple Queue Service",
+    "amazon_redshift": "Amazon Redshift",
+    "big_data_chip": "Big Data",
+    "change_data_capture_chip": "Change data capture",
+    "postgresDB": "postgresDB",
+    "sink_chip": "Sink",
+    "source_chip": "Source",
+    "streaming_messaging_chip": "Streaming&messaging",
+    "verison": "Version: 1.0.1 Final"
```
gashcrumb, line 33: Typo here, `version`

Kells512:
@gashcrumb I applied the comments you mentioned :)

src/ProofOfConcepts/connectors/connectors.tsx
```diff
@@ -0,0 +1,361 @@
...
+  DataListAction,
+  Flex,
+  FlexItem,
+  Label,
+  LabelGroup,
+} from "@patternfly/react-core";
+import PlusCircleIcon from "@patternfly/react-icons/dist/esm/icons/plus-circle-icon";
+import OutlinedQuestionCircleIcon from "@patternfly/react-icons/dist/esm/icons/outlined-question-circle-icon";
+
+export const ConnectorTypeListItem: React.FunctionComponent = () => {
```
gashcrumb, line 17: Maybe call this component the list, such as `ConnectorTypeListItem` and then it'd then contain `ConnectorTypeListItem`s.  It'd be worth refactoring out the `DataListItem` elements into it's own component that you can then pass the needed properties to.  Here's an example of what I mean, [this is the old component](https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/blob/main/src/app/components/ConnectorTypeCard/ConnectorTypeCard.tsx) that this POC could be replacing.

Kells512: Would you mean creating a props component and then also having the data list component separate?

gashcrumb: Sorry, I didn't catch this earlier :smile: 

See if you can take one of the `DataListItem` elements underneath this `DataList` you've created and turn it into it's own component that accepts properties for each of the items it needs to display.

gashcrumb:
I think we can close this as this effort was moved over to the cos-ui repo.

PR #584, opened 2022-12-01
chore: Refactor kafka topic

HemaHG:

> Refactor topic table


HemaHG:
@Riccardo as you were saying yesterday you handled the Kafka instances fetch in this repo which is not a good idea so you changed the approach, I am also doing the same thing in getting the kafka topics so I removed the data fetching part, please let me know if it's a good approach or should I stick with the existing code

riccardo-forina:
Yes it's a good idea, and a lot of that code feels like it can be handled as some shared hook of some sort, there is a lot of repetition between all the components that feed the tables with data.

PR #587, opened 2022-12-13
feat: Add analytics tracking to Data Science Page buttons.

cfchase:

Add segment.io analytics to track users button clicks on the Data Science Page.

**Verification steps** 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "@patternfly/react-core": "4.224.1",
         "@patternfly/react-icons": "4.75.1",
         "@patternfly/react-table": "4.93.1",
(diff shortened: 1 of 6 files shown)


riccardo-forina:
Hey @cfchase, thank you for this, but this is not going to work. 

This library is about visual components only. Nothing else can be done on this level.

If something needs to be tracked for analytics purposes, it must be done on the app using these components (in this case, app-services-ui).

It's acceptable to modify the components here to add support for tracking interactions, for example by adding a callback on click events that then the app can use to connect the tracking function.

Hope it makes sense.

cfchase:
paired with https://github.com/redhat-developer/app-services-ui/pull/726

riccardo-forina (approved):

PR #600, opened 2023-01-05
feat: UI for sample code snippet

hemahg:

> UI for the sample code snippet

 > fixes https://issues.redhat.com/browse/MGDSTRM-1432

**Verification steps** 

> https://619e1921b3e38b003afeb2be-phumbmbpjs.chromatic.com/?path=/story/kafka-kafkainstancedrawer-kafkasamplecode--sample-code-snippet


src/Kafka/KafkaInstanceDrawer/components/SampleCodeSelect.tsx
```diff
@@ -0,0 +1,505 @@
...
+        
+         public class KafkaConfig {
+        
+             public static Map<String, Object> Config() {
+        
+                 Map<String, Object> config = new HashMap<>();
+        
+                 String kafkaHost = System.getenv("<bootstrap_server>");
+                 String rhoasClientID = System.getenv("<client_id>");
+                 String rhoasClientSecret = System.getenv("{client_secret}");
```
jgiardino, line 317: This should have angle brackets (<>) instead of the curly brackets ({ })

src/Kafka/KafkaInstanceDrawer/components/SampleCodeSelect.tsx
```diff
@@ -0,0 +1,505 @@
...
+          import java.util.HashMap;
+         import java.util.Map;
+        
+         public class KafkaConfig {
+        
+             public static Map<String, Object> Config() {
+        
+                 Map<String, Object> config = new HashMap<>();
+        
+                 String kafkaHost = System.getenv("<bootstrap_server>");
```
jgiardino, line 315: For client ID and client secret, we want to use the angle brackets to communicate to the user that those are values the user needs to specify.

But for this line, we know what the bootstrap URL is for the selected Kafka instance, so we will need to prepopulate that in this sample code. But looking at the current UI preview with the description text, it looks like the user is expected to provide all 3 values. 

@riccardo-forina - since this bootstrap URL will be generated within consoledot based on the selected Kafka, is there another way we should represent this value in the storybook component so that it's more obvious it will be prepopulated? 

The following is an example of what the current UI component looks like.

image

jgiardino:
Thanks for sharing these updates, @hemahg! I'm adding @aprilma419 to also provide a review. 

Something that would be good is to define a max height on the code blocks to keep the panel a bit more manageable. It looks like the only way to do this is by using the expandable option shown in this example: https://www.patternfly.org/v4/components/code-block#expandable

It looks like we can control how many lines are visible before the rest are hidden in the expandable section. @aprilma419 or @riccardo-forina — do you have any suggestions on how many lines should display? Do you see any other options for controlling the height of these code blocks?

@aprilma419  - You can check out the UI preview by clicking the Details link for Storybook Publish under the checks section at the end of this PR:
image

Then navigate to the following storybook component in the storybook nav
image

aprilma419:
> Thanks for sharing these updates, @hemahg! I'm adding @aprilma419 to also provide a review.
> 
> Something that would be good is to define a max height on the code blocks to keep the panel a bit more manageable. It looks like the only way to do this is by using the expandable option shown in this example: https://www.patternfly.org/v4/components/code-block#expandable
> 
> It looks like we can control how many lines are visible before the rest are hidden in the expandable section. @aprilma419 or @riccardo-forina — do you have any suggestions on how many lines should display? Do you see any other options for controlling the height of these code blocks?
> 
> @aprilma419 - You can check out the UI preview by clicking the Details link for Storybook Publish under the checks section at the end of this PR: image
> 
> Then navigate to the following storybook component in the storybook nav image

@jgiardino Thanks, will take a look! My initial thought was to let the code block's height depend on its actual lines count, but considering not to make the page too long, using the expandable and having a default max height make sense to me.

src/Kafka/KafkaInstanceDrawer/CodeSnippets.ts
```diff
@@ -0,0 +1,420 @@
+export const javaConfig = `package org.example;
```
hemahg, line 1: @riccardo-forina could you please review the PR, I am treating the code snippet for each client as a const string and displaying that string in the code block, WDYT about this approach, are there any other ways I can try?

riccardo-forina: Yes it's fine

riccardo-forina:
I'd close this in favor of https://github.com/redhat-developer/rhosak-ui/pull/64 as I'd want to avoid releasing new things with the old codebase

PR #607, opened 2023-01-19
chore: Add a play function to correctly show invalid topic name story

suyash-naithani:
Added a play function that was missing in the resource name , invalid topic name story


riccardo-forina:
Since this component isn't in use in the legacy repo and we are moving to the new rhosak-ui one, please do the change only on the rhosak-ui repo.

PR #614, opened 2023-01-23
chore: Add checkbox for rows in responsive table

suyash-naithani:


riccardo-forina:
You are having issues because that's not composing the table the right way. Check the official example https://www.patternfly.org/v4/components/table#composable-selectable-with-checkbox
You need to add a label-less column at the beginning that is the one providing the checkbox. Note that this applies to the header row as well.
Then there is work to do in the area where the responsive logic is handled so as to not compute this column in the calculation.

hemahg (approved):
looks good to me

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -60,12 +61,16 @@ export type ResponsiveTableProps<TRow, TCol> = {
...
   isRowDeleted?: (props: RowProps<TRow>) => boolean;
   isRowSelected?: (props: RowProps<TRow>) => boolean;
+  isRowChecked?: (rowIndex: number) => boolean;
   expectedLength?: number;
   onRowClick?: (props: RowProps<TRow>) => void;
   setActionCellOuiaId?: (props: RowProps<TRow>) => string;
   setRowOuiaId?: (props: RowProps<TRow>) => string;
   tableOuiaId?: string;
   variant?: TableVariant;
+  onCheck?: (isSelecting: boolean, rowIndex: number) => void;
```
riccardo-forina, line 71: If you look at how the other callbacks are implemented, we pass the row object back to the callback, not an index. Eg. `isRowSelected?: (props: RowProps<TRow>) => boolean;`

This is because the index means nothing for who is using the table, as it depends on things like sorting and pagination that are too messy to deal with outside of the table itself.

Please correct it on this and the other new callbacks.

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -190,7 +201,19 @@ export const ResponsiveTable = <TRow, TCol>({
       variant={variant}
     >
       <Thead>
-        <Tr>{header}</Tr>
+        <Tr>
+          {onBulkSelect !== undefined && (
```
riccardo-forina, line 205: So if I pass `onBulkSelect` but on `onCheck` I get the checkbox on the header but not on the rows.

On the render function, we must check for the correctness of compound props, and `throw` an exception if they are not used as they should so to warn the developer about the mistake.

Also, name consistency please. If we used `onCheck`, then it should be `onBulkCheck`. We are using `select` for the action of highlighting a row, so this can be mistaken as a way to highlight all rows at once (which can't be done).

riccardo-forina (changes requested):

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -180,6 +189,31 @@ export const ResponsiveTable = <TRow, TCol>({
...
 
+  const allChecked = areAllRowsChecked != undefined && areAllRowsChecked();
+
+  const isBulkCheckPropsValid = () => {
+    if (onBulkCheck != undefined && areAllRowsChecked != undefined) return true;
+    else if (
+      (onBulkCheck != undefined && areAllRowsChecked == undefined) ||
+      (onBulkCheck == undefined && areAllRowsChecked != undefined)
+    )
+      console.warn(`Not all props for checkbox have been defined`);
```
riccardo-forina: I asked for a `throw`, not a log

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -180,6 +189,31 @@ export const ResponsiveTable = <TRow, TCol>({
...
+    return false;
+  };
+
+  const isRowCheckPropsValid = () => {
+    if (isRowChecked != undefined && onCheck != undefined) return true;
+    else if (
+      (isRowChecked != undefined && onCheck == undefined) ||
+      (isRowChecked == undefined && onCheck != undefined)
+    )
+      console.warn(`Not all props for checkbox have been defined`);
```
riccardo-forina: I asked for a `throw`, not a log

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -180,6 +189,31 @@ export const ResponsiveTable = <TRow, TCol>({
...
+    else if (
+      (isRowChecked != undefined && onCheck == undefined) ||
+      (isRowChecked == undefined && onCheck != undefined)
+    )
+      throw new Error("Not all props for checkbox have been defined");
+    return false;
+  };
+
+  const bulkCheckPropsValidation = isBulkCheckPropsValid();
+  const rowCheckPropsValidation = isRowCheckPropsValid();
```
riccardo-forina, line 215: If you throw, this code can't be reached, making this boolean and its usage meaningless.

src/shared/Table/ResponsiveTable.tsx
```diff
@@ -180,6 +189,31 @@ export const ResponsiveTable = <TRow, TCol>({
...
+    return false;
+  };
+
+  const isRowCheckPropsValid = () => {
+    if (isRowChecked != undefined && onCheck != undefined) return true;
+    else if (
+      (isRowChecked != undefined && onCheck == undefined) ||
+      (isRowChecked == undefined && onCheck != undefined)
+    )
+      throw new Error("Not all props for checkbox have been defined");
```
riccardo-forina, line 210: The messages can be more meaningful. One shouldn't have to jump into the source code to understand what the error is exactly.

riccardo-forina (changes requested):
I'd suggest stopping for a moment to understand the implications behind a suggestion before committing and re-asking for a review.

suyash-naithani:
I've made the message more meaningful, let me know if that is better. I am also catching the error being thrown and logging it as an error in the console so the code does no stop executing. Let me know what you think

riccardo-forina:
Stopping the code from being executed was exactly the point of throwing instead of logging. If the developer does a mistake in using the table, we must let him know asap in the most immediate way possible. Throwing will break the app, or trigger any error boundary in place, making the dev immediately aware that something is amiss.

suyash-naithani:
So if I throw the error, the app breaks , you said that the ` const bulkCheckPropsValidation = isBulkCheckPropsValid()` code cannot be reached when I throw the error, making this boolean and it's usage meaningless which is why I added the console error. Did you mean using the `throw` somewhere else than where it is currently used

riccardo-forina:
No, just that the boolean return and checking for it is meaningless

riccardo-forina (approved):

PR #617, opened 2023-02-09
chore: Connector Message Statisics Drawer

Kells512:

> Be able show what https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/pull/733 looks like in storybook 

> To show what the examples for the message drawer look like

**Verification steps** 

> All examples found in the POC - CONNECTORS MESSAGE DRAWER

>Was not able to exactly move over the code from cos-ui to here, so open another pr there with what the code part should look like https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/pull/733


gashcrumb:
@Kells512 I think what you've done on this branch is a good direction.  I'd suggest just using the Drawer component in the stories just to wrap each story in that and let the components that you're creating focus on the contents.  So for example, were I to integrate some of these components, I'd probably just use `Drawer` directly and would want to have header and drawer content components available separately when composing the page.

> Was not able to exactly move over the code from cos-ui to here, so open another pr there with what the code part should look like https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/pull/733

Which part(s), can you maybe comment on the lines in that other PR?  Happy to help figure out how to bring everything over that needs to come over.

Kells512:
@gashcrumb It was just the other files that would have shown the message statistics draw on the whole page, I was not able to get some of the imports to work. And also the translation strings were also causing issues, I commented on the ones in the second PR

gashcrumb:
@Kells512 ah, okay gotcha, thanks!  I'll take a closer look across these two sets of changes, probably there's a number of components in there that wouldn't be in this repo.  My initial thought is maybe just mock up anything not to do with message statistics in the Story so that we can see how the component will look when used.  I'd be happy to go and scope this out, I'm not sure without looking if the components that make up that drawer hide a lot of nested markup that'd be tedious to mock.

Kells512:
@gashcrumb That would be really helpful, thanks Stan.

gashcrumb:
One thing I probably should have pointed out earlier, you'll want to replace `varient` with `variant` in this PR.

gashcrumb:
So, having a look and comparing the two, I think I'd let this stay focused on just adding the message statistics components, but rather than use CSS to get things positioned, I'd wrap each story in a generic drawer much like you've commented out [here](https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/pull/733/files#diff-537639c3d99d2a2db7bb39dff07fead3d99d7e3011daa75efee9472a93f16666R70-R79).  I think since each of these additions in the design are in their own row just seeing them in the context of the drawer should be enough.

If we do want some surrounding connectors content, I'm kind of wondering if it wouldn't be nice to have some general placeholder component that uses Skeleton to just mock the content around the things we're interested in.  But let's worry about that when we need to :smile:

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerDescList/ConnectorDrawer.css
```diff
@@ -0,0 +1,31 @@
...
+}
+.connector-drawer__header-title {
+  margin-top: 0 !important;
+}
+.connector-drawer__tab-content {
+  padding-top: var(--pf-global--spacer--lg);
+  padding-bottom: var(--pf-global--spacer--lg);
+}
+
+.pf-u-icon-color-green {
```
gashcrumb, line 12: If you use the `@patternfly/react-icons` components, you can use the `color` property to set this color, which I think would be better than doing this via CSS.  [Like this](https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/blob/main/src/app/components/ConnectorTypeCard/ConnectorTypeCard.tsx#L58) for example.  The constant I'm using there is from `@patternfly/react-tokens` which contains similar variables that [you can import](https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/blob/main/src/app/components/ConnectorTypeCard/ConnectorTypeCard.tsx#L22) and use.

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerGrid/ConnectorDrawerMessageStatistics.tsx
```diff
@@ -0,0 +1,91 @@
...
+  GridItem,
+} from "@patternfly/react-core";
+import { ExclamationIcon } from "@patternfly/react-icons";
+import CheckIcon from "@patternfly/react-icons/dist/esm/icons/check-icon";
+
+import "./ConnectorDrawer.css";
+
+export const ConnectorDrawerMessageStatistics = () => {
+  return (
+    <Flex direction={{ default: "column" }}>
```
gashcrumb, line 30: Wonder, why Flex and not [Stack](https://patternfly-react.surge.sh/layouts/stack) out of curiousity?

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerDescList/ConnectorDrawer.css
riccardo-forina: For what is worth, when migrating all the Kafka components to the new rhosak-ui repo I removed _all_ CSS files and replaced them all with PF utility classes where needed.

gashcrumb: Yep, totally agree.  For what we're working with it's really better to avoid needing CSS to lay out components, it's ideally handled by a combination of the components, their properties and the [utility classes](https://www.patternfly.org/v4/utilities/alignment).  And then when all that fails and you need that hammer :hammer:, you pull out the CSS :smile:

Kells512: I went thought the CSS rules and removed anything that was not relevant for the pr for now,  the only thing left was a CSS rule to make text grey

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerGrid/ConnectorDrawerMessageStatistics.tsx
Kells512: I actually had not considered stack, I started with flex and it seemed to be working

riccardo-forina:
Just a quick note to start considering moving these project-specific POCs to the project's own repo, as ui-components will eventually be reduced to just the shared components and will not contain the project-specific ones anymore.

jgiardino:
Thanks for the update, @riccardo-forina. How much time would you say we have with this PR? I think we're close to finalizing the POCs for review with the rest of the team.

riccardo-forina:
Nothing is planned yet, and there is no rush. Ideally after Hybrid is live we'll start some cleanup tasks.

jgiardino:
Your latest updates look really good! Here are some quick comments from my latest review

**Variant 4**
When using the description list component for the metrics, you have a title component included for the `<DescriptionListTerm>` but instead we should just use the oob styles provided by `<DescriptionListTerm>` and not include the additional `<Title>` component. Here is a side-by-side comparison, where the first metric uses the oob styles.

image

I think for all of the examples, the icon should be included as part of the `<DescriptionListDescription>` and the size should be controlled by the modifier class applied to the parent component. This should render the icon at the same font size as the text it precedes. Right now, it's really small relative to the text that accompanies it.


I think you could go ahead and drop variant 3. I don't think the layout works well at presenting the contents needed for these metrics (e.g. the description text).

Kells512:
@jgiardino I think this is ready to be viewed by others

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerSplit/ConnectorDrawerMessageStatistics.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <CardTitle>
+            <Title headingLevel="h4">Processed messages</Title>
+          </CardTitle>
+          <CardBody>
+            <Split>
+              <SplitItem isFilled>
+                {/** displaysize in description list will be better use then utilites classes, but throwing errors for some reason */}
+                <Flex
+                  className="pf-u-font-size-2xl"
+                  direction={{ default: "row" }}
```
jgiardino, line 58: I don't think the Flex layout includes a prop to prevent wrapping, so I think you also need to include the utility class `pf-u-flex-nowrap` on the flex layouts here. This keeps the first line of text in the same line as the icon, and forces only the text to wrap, rather than moving the whole text block underneath the icon.

src/ProofOfConcepts/ConnectorMessageDrawer/DrawerSplit/ConnectorDrawerMessageStatistics.tsx
```diff
@@ -0,0 +1,113 @@
...
+> = ({ sent, notSent }) => {
+  return (
+    <Flex direction={{ default: "column" }}>
+      <FlexItem>
+        <Card>
+          <CardTitle>
+            <Title headingLevel="h4">Processed messages</Title>
+          </CardTitle>
+          <CardBody>
+            <Split>
```
jgiardino, line 53: The Split layout supports including a gutter, and we should use that here. 

In theory the gutter size here should be larger than the space between the icon and the text, so that "1600 sent" is never closer to the "!" icon vs the ✅ icon. But this would require that the flex layout used below has a `spaceitems` prop that sets the space to `sm` or smaller. But I don't think that works with the current font size at 2xl. The other alternative is to replace the Split layout with a flex layout where we have more control over the space between flex items, so we could have a `lg` space between the 2 metrics and a `md` space between the icon and text. 🤷‍♀️  

image

PR #618, opened 2023-02-16
chore: Add kebab in the toolbar

suyash-naithani:

> We have the option to show toolbar in responsive table but it was not showing up because of a missing prop.
> Ad the missing prop and expand on the type of `ToolbarAction` to allow sending dropdown values and to enable/disable the dropdown items

### DIFF
--- a/src/shared/Table/TableView.tsx
+++ b/src/shared/Table/TableView.tsx
@@ -44,7 +44,6 @@ export type ToolbarAction = {
   label: string;
   isPrimary: boolean;
   onClick: () => void;
-  hasKebabToolbarAction?: boolean; //Tells us if the action item is a kebab dropdown
   dropdownItems?: ActionDropdownItemsType[];
 };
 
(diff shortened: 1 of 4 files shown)


src/shared/Table/TableView.tsx
```diff
@@ -198,37 +208,52 @@ export const TableView = <TRow, TCol>({
...
-                      }}
-                    >
-                      {a.label}
-                    </OverflowMenuDropdownItem>
-                  ))}
-                  isFlipEnabled
-                  menuAppendTo="parent"
-                />
+              <OverflowMenuControl
+                hasAdditionalOptions={isToolbarKebabActionVisible}
```
hemahg, line 227: I guess to can use actions.some((a) => a.dropdownItems) to check hasAdditionalOptions props instead of passing a prop

src/shared/Table/TableView.tsx
```diff
@@ -96,6 +104,39 @@ export const TableView = <TRow, TCol>({
...
+        }}
+        isDisabled={a.isDisabled}
+      >
+        {a.label}
+      </OverflowMenuDropdownItem>
+    ));
+    if (buttonActions != undefined && kebabActionItems != undefined)
+      return buttonActions.concat(kebabActionItems);
+    else if (buttonActions != undefined) return buttonActions;
+    return kebabActionItems;
```
riccardo-forina, line 136:
```suggestion
    if (buttonActions != undefined && kebabActionItems != undefined)
      return buttonActions.concat(kebabActionItems);
    else if (buttonActions != undefined) return buttonActions;
    return [
       ...(buttonActions || []),
       ...(kebabActionItems || []),
    ];
```

`concat` mutates the original array in place. I'd suggest avoiding using it unless there is a good reason for not doing so.

PR #619, opened 2023-02-16
chore: prevent duplicate keys in merged array and remove condition fo…

suyash-naithani:
- When both action button and kebab are present, merging the two in a dropdown can cause duplicate keys and a warning to show up in the console. Prefix the kebab items key with 'kebab' to avoid duplicate keys
- We do not need a condition to check for accounts, it does not show any visible white space. In the condition where we will have only kebab options and no action buttons, the condition will prevent us from seeing the kebab. Removed the condition.

### DIFF
--- a/src/shared/Table/TableView.tsx
+++ b/src/shared/Table/TableView.tsx
@@ -110,7 +110,7 @@ export const TableView = <TRow, TCol>({
       <OverflowMenuDropdownItem
         key={id}
         onClick={() => {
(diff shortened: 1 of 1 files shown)


suyash-naithani:
@riccardo-forina  I noticed some changes needed in the action items kebab and addressed them here, please refer to the description. Apologies for not noticing this earlier

riccardo-forina (approved):

PR #622, opened 2023-03-02
chore: Get row data when a row is checked

suyash-naithani:


src/shared/Table/ResponsiveTable.tsx
```diff
@@ -68,7 +68,7 @@ export type ResponsiveTableProps<TRow, TCol> = {
   setRowOuiaId?: (props: RowProps<TRow>) => string;
   tableOuiaId?: string;
   variant?: TableVariant;
-  onCheck?: (isSelecting: boolean, rowIndex: number) => void;
+  onCheck?: (isSelecting: boolean, row: RowProps<TRow>) => void;
```
riccardo-forina, line 71: The row param is always the first one in the other callbacks. I'd do the same here for consistency

riccardo-forina (approved):
Left a comment, it would be nice if you could address it before merging

PR #626, opened 2023-03-28
chore: Add aria-label props for toolbar

suyash-naithani:

>Add props for aria-label for table toolbar and kebab options

### DIFF
--- a/src/shared/Table/TableView.stories.tsx
+++ b/src/shared/Table/TableView.stories.tsx
@@ -102,6 +102,8 @@ Example.args = {
 export const KebabAndButtonToolbarVisible = Template.bind({});
 KebabAndButtonToolbarVisible.args = {
   data: sampleData,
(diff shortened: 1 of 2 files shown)


hemahg (approved):

src/shared/Table/TableView.stories.tsx
```diff
@@ -102,6 +102,8 @@ Example.args = {
 export const KebabAndButtonToolbarVisible = Template.bind({});
 KebabAndButtonToolbarVisible.args = {
   data: sampleData,
+  toolbarKebabAriaLabel: "kebab-dropdown-toolbar",
+  toolbarAriaLabel: "table-toolbar",
   kebabActions: [
```
jgiardino, line 107: Aria labels are user-facing contents and are not intended to provide IDs on HTML elements. These need to be human-readable strings.

